PR #12, opened 2021-07-13
Publish binary releases

aalmiray:
Fixes https://github.com/gunnarmorling/kcctl/issues/10

- Add Maven configuration for creating Native Image distributions -> https://jreleaser.org/guide/latest/distributions/native-image.html
- Build main branch on every push.
- Publish rolling releases only when pushing to `main`.


gunnarmorling (approved):

gunnarmorling:
Applied, let's see what happens :)

PR #16, opened 2021-07-15
🌟 Configure release workflow

aalmiray:
This PR configures a release workflow for stable versions.

A release must be triggered manually from the workflows tab. There are 2 inputs:
* the version to be released, i.e, `1.0.0.Alpha1`. (REQUIRED)
* the next development version, i.e, `1.1.0-SNAPSHOT`. (OPTIONAL)

The next development version will be automatically calculated from the current version, that is, use the `d.d.d` value and append `-SNAPSHOT`. No number will be automatically incremented as the `maven-release-plugin` does because one could go from `1.0.0.Final` into `1.1.0-SNAPSHOT` (or even `2.0.0-SNAPSHOT` !!) where as the maven release plugin will always increment the build number, resulting in `1.0.1-SNAPSHOT` for example instead of `1.1.0-SNAPSHOT`. Also, the release plugin would require using the `jreleaser-maven-plugin` and executing the `jreleaser:full-release` goal.

This workflow could be revised later.

### DIFF
--- a/.github/workflows/early-access.yml
+++ b/.github/workflows/early-access.yml
@@ -24,7 +24,7 @@ jobs:
   # Build native executable per runner
   build:
     name: 'Build with Graal on ${{ matrix.os }}'
-    if: github.repository == 'gunnarmorling/kcctl' && startsWith(github.event.head_commit.message, 'Releasing version') != true
+    if: github.repository == 'gunnarmorling/kcctl' && startsWith(github.event.head_commit.message, '🏁 Releasing version') != true && startsWith(github.event.head_commit.message, '⬆️  Next version') != true
     strategy:
       fail-fast: true
       matrix:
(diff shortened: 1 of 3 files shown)


.github/workflows/early-access.yml
```diff
@@ -24,7 +24,7 @@ jobs:
   # Build native executable per runner
   build:
     name: 'Build with Graal on ${{ matrix.os }}'
-    if: github.repository == 'gunnarmorling/kcctl' && startsWith(github.event.head_commit.message, 'Releasing version') != true
+    if: github.repository == 'gunnarmorling/kcctl' && startsWith(github.event.head_commit.message, '🏁 Releasing version') != true && startsWith(github.event.head_commit.message, '⬆️  Next version') != true
```
gunnarmorling, line 27: I like your attention to details ;)

.github/workflows/release.yml
```diff
@@ -0,0 +1,213 @@
...
+          path: ~/.m2
+          key: ${{ runner.os }}-m2-${{ hashFiles('**/pom.xml') }}
+          restore-keys: ${{ runner.os }}-m2
+          
+      - name: 'Set next version'
+        env:
+          NEXT_VERSION: ${{ steps.version.outputs.NEXT_VERSION }}
+        run: |
+          ./mvnw -B versions:set versions:commit -DnewVersion=$NEXT_VERSION
+          git config --global user.email "41898282+github-actions[bot]@users.noreply.github.com"
```
gunnarmorling, line 210: What's that number?

gunnarmorling (approved):
LGTM; thanks a lot! Feel free to merge yourself when you deem it done?

.github/workflows/release.yml
aalmiray: It's a magic number for the github-bot account

gunnarmorling:
Thanks!

>

Issue #26, opened 2021-07-28
Configure JReleaser as a Maven plugin


aalmiray:
This will let anyone test out the release configuration without having the CLI pre-installed.

Requires JReleaser `0.6.0` as a minimum.

Must update release workflows as well.

PR #27, opened 2021-07-28
🦉 Configure JReleaser via its Maven plugin

aalmiray:
Fix for #26


aalmiray:
Converting to draft as JReleaser `0.6.0` is needed before this PR can be merged.

pom.xml
```diff
@@ -306,5 +319,49 @@
...
+                <distributions>
+                  <kcctl>
+                    <name>kcctl</name>
+                    <type>NATIVE_IMAGE</type>
+                    <artifacts>
+                      <artifact>
+                        <path>{{artifactsDir}}/{{distributionName}}-{{projectVersion}}-linux-x86_64.zip</path>
+                        <transform>artifacts/{{distributionName}}-{{projectEffectiveVersion}}-linux-x86_64.zip</transform>
+                        <platform>linux-x86_64</platform>
+                      </artifact>
```
gunnarmorling, line 346: Isn't this missing the TAR.GZ flavour?

aalmiray: Yes. Are we producing one?

aalmiray: According to `src/main/assembly/assembly.xml` we only produce ZIP files for distributions.

gunnarmorling: Hum, that's really weird. You are right, but I could swear I configured it recently to produce both TAR.GZ and ZIP.

aalmiray: Wans't that the case for kcetcd?

gunnarmorling: Ah yes, of course. Ok, so at least I'm not dreaming :) Could you update the PR so we create and publish both formats?

aalmiray: OK, so you want kcctl to publish 6 artifacts in total?

gunnarmorling:
> OK, so you want kcctl to publish 6 artifacts in total?

Yeah, or do you believe that's overkill? I always like leaving people the
choice between the archive format they prefer. E.g. imposing ZIP on Linux
users seems a bit odd. WDYT?

>

aalmiray:
No, not really. However the inverse is true, l: offer tar.gz for the other platforms. Even though the build can produce 6 distro files only 3 may be released: 2 zips, 1 tar. WDYT?

gunnarmorling:
> 2 zips, 1 tar. WDYT?

So ZIP for Windows and Mac, and TAR.GZ for Linux? That'd work for me, too.

aalmiray:
OK. Will send another PR first that targets the current state. Then rebase this PR after merge.

aalmiray:
Depends on https://github.com/kcctl/kcctl/pull/29

pom.xml
```diff
@@ -36,11 +36,24 @@
...
+  <description>kcctl -- A CLI for Apache Kafka Connect</description>
+  <inceptionYear>2021</inceptionYear>
+  <url>https://github.com/kcctl/kcctl</url>
+
+  <developers>
+    <developer>
+      <id>gunnarmorling</id>
+      <name>Gunnar Morling</name>
+    </developer>
+  </developers>
```
gunnarmorling, line 52: Wanna add yourself too? Seems more than justified :)

gunnarmorling (approved):

gunnarmorling:
Applied, thx! If you want to add yourself as a developer as suggested in my last comment, just push a separate commit for that to main.

PR #29, opened 2021-07-29
🤐 Use tar.gz for linux distribution instead of .zip

aalmiray:


gunnarmorling (approved):

gunnarmorling:
Applied; thx, @aalmiray!

PR #30, opened 2021-08-04
🍷 Update Graal to 21.2.0

aalmiray:
Follow up to https://twitter.com/gunnarmorling/status/1423022723350122497?s=20

I've got a small reduction of the binary on my local environment, from 44M to 41M.


gunnarmorling:
Nice, thx! Observing 41 MB -> 39 MB on macOS locally.

PR #31, opened 2021-08-06
Add describe tasks-config command

mimaison:
This maps to /connectors/<connector>/tasks-config and returns the configuration of all tasks of a connector.
Usage: kcctl describe tasks-config <connector>

### DIFF
--- a/src/main/java/dev/morling/kccli/command/DescribeConnectorCommand.java
+++ b/src/main/java/dev/morling/kccli/command/DescribeConnectorCommand.java
@@ -31,7 +31,10 @@ import dev.morling.kccli.service.KafkaConnectApi;
 import dev.morling.kccli.service.TaskState;
 import dev.morling.kccli.service.TopicsInfo;
 import dev.morling.kccli.util.ConfigurationContext;
+import dev.morling.kccli.util.Tuple;
+import dev.morling.kccli.util.Version;
 import picocli.CommandLine.Command;
+import picocli.CommandLine.Option;
 import picocli.CommandLine.Parameters;
 
 import static dev.morling.kccli.util.Colors.ANSI_GREEN;
@@ -48,11 +51,24 @@ public class DescribeConnectorCommand implements Runnable {
     @Parameters(paramLabel = "CONNECTOR NAME", description = "Name of the connector") // , completionCandidates = DummyCompletions.class)
     String name;
 
+    @Option(names = { "--tasks-config" }, description = "Displays tasks configuration")
+    boolean includeTasksConfig;
+
+    private final Version requiredVersionForTasksConfig = new Version(2, 8);
+
     @Override
     public void run() {
         KafkaConnectApi kafkaConnectApi = RestClientBuilder.newBuilder()
                 .baseUri(context.getCluster())
                 .build(KafkaConnectApi.class);
+
+        if (includeTasksConfig) {
+            Version currentVersion = new Version(kafkaConnectApi.getWorkerInfo().version);
+            if (!currentVersion.greaterOrEquals(requiredVersionForTasksConfig)) {
+                System.out.println("--tasks-config requires at least Kafka Connect 2.8. Current version: " + currentVersion);
+                return;
+            }
+        }
 
         ConnectorInfo connector = kafkaConnectApi.getConnector(name);
         ConnectorStatusInfo connectorStatus = kafkaConnectApi.getConnectorStatus(name);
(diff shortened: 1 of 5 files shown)


gunnarmorling:
Hey @mimaison, welcome, and thanks a lot for this contribution! I think it's a great addition, two questions only on the way of delivering this feature:

* This end-point is only available in AK 2.8 and newer, invoking the command against an earlier version currently prints an ugly stack trace due to the 404 returned by the REST API. Should we query the version (it already is exposed via `kcctl info`)  and display a nice error message in case the API doesn't exist?
* Should this be its own sub-command under `describe` indeed, or should we rather show the tasks config upon `describe connector <connector id>`? Thinking about it, this would allow for a transparent handling of the first issue, as we simply would or would not show the tasks config based on the backend's capabilities. Note that `describe` already yields an aggregated view, i.e. there's no 1:1 relationship between API methods and kcctl commands. Or do you think this would be too much info to show for a single command?

gunnarmorling:
Hey @mimaison, howdy? Any thoughts on the questions above?

mimaison:
Hi @gunnarmorling

Thanks for the feedback, I'll hopefully have the time to update the PR tomorrow.

- You're right, this endpoint is still very new. Many users probably don't run 2.8 yet so it makes sense to check compatibility and handle older versions gracefully.
- I wondered of the best way to expose this feature. I think task-configs is a lot of details not all users may want to see all the time. Also with many tasks it can be very verbose! So I wasn't sure if adding it to `describe` was the best option, that's why I added a new command. Another option is to add an extra flag to `describe` to get all the details. WDYT?

gunnarmorling:
Good point re verbosity. In that light, yes, adding a flag to `describe` would be my preference, I think. It feels more consistent in a way.

mimaison:
Thanks @gunnarmorling 
I've pushed an update. Let me know if you have any feedback.

src/main/java/dev/morling/kccli/command/DescribeConnectorCommand.java
```diff
@@ -85,17 +93,37 @@ public void run() {
...
-        printTuples(Arrays.asList(new Tuple(ANSI_WHITE_BOLD + "Topics" + ANSI_RESET, "")));
+        Tuple.print(Arrays.asList(new Tuple(ANSI_WHITE_BOLD + "Topics" + ANSI_RESET, "")));
 
         List<Tuple> topics = new ArrayList<>();
 
         for (String topic : connectorTopics.entrySet().iterator().next().getValue().topics) {
             topics.add(new Tuple("", "  " + topic));
         }
-        printTuples(topics);
+        Tuple.print(topics);
```
gunnarmorling, line 106: Let's put the tasks config above topics. That way the configs (connector + tasks) is next to each other which I feel is more consistent.

src/main/java/dev/morling/kccli/command/DescribeConnectorCommand.java
```diff
@@ -85,17 +93,37 @@ public void run() {
...
+                for (Entry<String, Map<String, String>> task : tasksConfigs.entrySet()) {
+                    tuples.add(new Tuple("  Task", task.getKey()));
+                    for (Entry<String, String> taskConfig : task.getValue().entrySet()) {
+                        tuples.add(new Tuple("    " + taskConfig.getKey(), taskConfig.getValue()));
+                    }
+                }
+                Tuple.print(tuples);
+            }
+            else {
+                System.out.println("  --tasks-config requires at least Kafka Connect 2.8. Current version: " + currentVersion);
```
gunnarmorling, line 124: So do you think we actually should show the other output in this situation; or should we rather handle this as sort of an invocation failure more explicitly and _only_ show this message when passing the parameter with KC < 2.8?

gunnarmorling (approved):
This looks mostly good now, @mimaison! One more suggestion and one question inline. Besides that, could you also

* Log an issue for this feature
* Reword the commit message so it begins with the issue key and an emoji of your liking: "#123 🔎 Adding support for ..."; there shall be no emoji-less commit in this repo ;)

Thanks again!

mimaison:
Thanks @gunnarmorling!
I've addressed your suggestions.

gunnarmorling:
Excellent, applied. Thank you so much, @mimaison! New binaries with this change will be available automatially in a few min.

gunnarmorling:
Pushed one more commit for updating the completion script. That's a bit tedious atm., looking into making this process a bit smoother currently.

Issue #1, opened 2021-07-11
Add support for pause/resume


gunnarmorling:
As requested by @hpgrahsl, the following should work:

    kcctl pause my-connector/0
    kcctl resume my-connector/0

gunnarmorling:
Done. Output coloring will be done via #5.

PR #35, opened 2021-08-13
Add pause and resume connector commands

karim-elngr:
Add two new commands to support pause and resume kafka connectors #1 

1- `kcctl pause connector connector-name-1`
2- `kcctl resume connector connector-name-1`

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,6 +43,8 @@ Commands:
   apply     Applies the given file for registering or updating a connector
   patch     Modifies the configuration of a connector or logger
   restart   Restarts a connector or task
+  pause     Pauses a connector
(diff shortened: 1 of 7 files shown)


gunnarmorling:
Thanks, that's a great start! Could you:

* rework the first commit using your e-mail linked to your GH account (as is the case with the second one), so to not have the "Octocat" commits
* run `mvn license:format` for adding the missing license headers
* update the output in _README.md_ from running `kcctl` without any options

Thanks!

gunnarmorling (approved):

gunnarmorling:
Great work, thanks a lot, @karim-elngr!

Issue #36, opened 2021-08-13
Allow to update connector configuration via patch


gunnarmorling:
E.g. like this:

    kcctl patch connector etcd-test -p '{"tasks.max":3}'
    kcctl patch connector etcd-test -p '{"tasks.max":3, "database.name":"foo"}'

gunnarmorling:
@SiddiqueAhmad, @karim-elngr, just in case you're looking for more things to pick up, this one could be an interesting one. It's about updating single properties of an already registered connector, instead of having to modify and apply the entire JSON file. Just an idea, please don't feel obligated in any way.

karim-elngr:
@gunnarmorling Awesome I would love to take care of this one

gunnarmorling:
It's all yours, thanks a lot :)

>

karim-elngr:
I was trying to understand how to get this to work, and I think we can do the following
 
1- Get the configuration parameters via the REST API
2- Update the configuration parameters using the supplied argument
3- Send the new configuration parameters via the REST API

The configuration parameters argument can be used in two ways

1- A `String` as in the example above `-p '{"tasks.max":3, "database.name":"foo"}'`
2- A `Map<String, String>` and use picocli to construct the map as follows `-p tasks.max=3 -p database.name=foo`

My own personal preference is to use the 2nd approach as it is avoids the user having to deal with JSON strings, and we can do the following to achieve the patch config:

```Java
// Declare the patch parameters
@Option(names = { "-p", "--parameter" }, description = "Configuration parameter for the connector")
Map<String, String> patchParameters;

// Use the patch parameters to update the configuration map
Map<String, String> connectorParameters = kafkaConnectApi.getConnectorConfig(name);
connectorParameters.putAll(patchParameters);

// Send the new configuration map to update the connector
String connectorParametersString = new ObjectMapper().writeValueAsString(connectorParameters);
kafkaConnectApi.updateConnector(name, connectorParametersString);
```

And finally we can print the new configurations for the specified connector using a describe connector command.

What do you think?

gunnarmorling:
Hey @karim-elngr, thanks for this great analysis!

> I think we can do the following

+1

> The configuration parameters argument can be used in two ways
> ...
> My own personal preference is to use the 2nd approach as it is avoids the user having to deal with JSON strings

[...]

PR #46, opened 2021-08-16
🩹 Add a command to patch connector parameters

karim-elngr:
#36

### DIFF
--- a/src/main/java/dev/morling/kccli/command/PatchConnectorCommand.java
+++ b/src/main/java/dev/morling/kccli/command/PatchConnectorCommand.java
@@ -1,0 +1,66 @@
+/*
+ *  Copyright 2021 The original authors
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package dev.morling.kccli.command;
+
+import java.util.Map;
+import java.util.concurrent.Callable;
+
+import javax.inject.Inject;
+
+import org.eclipse.microprofile.rest.client.RestClientBuilder;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+import dev.morling.kccli.service.KafkaConnectApi;
+import dev.morling.kccli.util.ConfigurationContext;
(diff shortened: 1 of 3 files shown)


src/main/java/dev/morling/kccli/command/PatchConnectorCommand.java
```diff
@@ -0,0 +1,66 @@
...
+
+        String connectorParametersString = new ObjectMapper().writeValueAsString(connectorParameters);
+        kafkaConnectApi.updateConnector(name, connectorParametersString);
+
+        DescribeConnectorCommand describeConnectorCommand = new DescribeConnectorCommand();
+        describeConnectorCommand.context = context;
+        describeConnectorCommand.name = name;
+        describeConnectorCommand.includeTasksConfig = false;
+
+        return describeConnectorCommand.call();
```
gunnarmorling, line 64: Perhaps print out a statement here such as "New connector configuration:"?

gunnarmorling (approved):
Very nice! One small suggestion inline. How does unsetting a connector option work? Can I do `-p foo=null`? Or maybe we should have both `-s` ("set") and `-r` ("remove")? WDYT?

Besides that, could add a completion which returns the already existing options (I've just pushed a commit which makes completion handing quite a bit simpler)? Then users will have <TAB> completion when updating an option, and will need to fully type the name only for adding a new connector config option.

src/main/java/dev/morling/kccli/command/PatchConnectorCommand.java
karim-elngr: Yes, let me add it

karim-elngr:
> Very nice! One small suggestion inline. How does unsetting a connector option work? Can I do `-p foo=null`? Or maybe we should have both `-s` ("set") and `-r` ("remove")? WDYT?

I thought so as well, we can introduce the null but we will have to parse the `"null"` string and remove the config from the map, alternatives I like the 2nd approach but perhaps we can omit the `-s` and default to adding a config unless `-r` is specified, WDYT?

karim-elngr:
> Besides that, could add a completion which returns the already existing options (I've just pushed a commit which makes completion handing quite a bit simpler)? Then users will have completion when updating an option, and will need to fully type the name only for adding a new connector config option.

I could not find any hints as to how to make auto-completion work based on a previous argument since we need to load configuration properties based on the connector name

gunnarmorling:
> I think I prefer the 2nd option because it seems less ambiguous, WDYT?

+1

> I could not find any hints as to how to make auto-completion work based on a previous argument since we need to load configuration properties based on the connector name

[This article](https://iridakos.com/programming/2018/03/01/bash-programmable-completion-tutorial) could be helpful; e.g. see `COMP_WORDS` and the related env vars which provide access to details of the current command line. You'd then have to add a parameter to the definition of the command invoked by the completion.

karim-elngr:
Hello @gunnarmorling I did not yet get to finish reading it, but I do not mind merging the PR and add a follow up to add the auto completion

gunnarmorling (approved):

gunnarmorling:
Ok, cool. PR merged. Thx a lot, @karim-elngr!

PR #48, opened 2021-08-18
Add support for multiple configuration contexts

### LINKED ISSUE #24: Expand notion of configuration context
In order to support display (#2) and reset (#8) of connector offsets, the concept of the configuration context should be expanded. We need to have the information about the Kafka bootstrap server(s) as well as the name of the offset topic. The suggestion is to introduce a notion of _named context_ which take all the required configuration and allowing the the user to switch between them (#19).

    # local is the name of the context
    $ kcctl config set-context local --cluster=my-connect:8083 --bootstrap-servers=my-kafka:9092 --offset-topic=my-offsets
    Using context local

The information about all the configured contexts would be stored in JSON format in the _~/.kcctl_ configuration file. Commands that don't need the additional attributes, should gracefully handle the current format of _~/.kcctl_ (solely containing the KC URL). Commands which do need the additional attributes, should raise an error when encountering a file in the legacy format, indicating how to use `set-context` for updating it.

helpermethod:
Hi @gunnarmorling,

here comes the PR! I hope I've got the the requirements right 😄.

One thing I would like to challenge is supporting the old, properties-like config format. Because we are still pre 1.0.0, I guess it would be acceptable to not support the current (properties-like) configuration format anymore. Wdyt?

The current, context-aware format is very flexible, except for the context-name and cluster URL, every other property is
optional.

So both

```json
{
    "currentContext": "local",
    "local": {
        "cluster": "http://localhost:8083"
    }
}
```

and

```json
{
    "currentContext": "local",
    "local": {
        "cluster": "http://localhost:8083",
        "boostrapServers": "localhost:9092",
        "offsetTopic": "connect-offsets"
    }
}
```

are valid.

Fixes #24.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -97,6 +97,17 @@
       <artifactId>quarkus-junit5</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.assertj</groupId>
+      <artifactId>assertj-core</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>net.javacrumbs.json-unit</groupId>
+      <artifactId>json-unit-assertj</artifactId>
+      <version>2.27.0</version>
+      <scope>test</scope>
+    </dependency>
   </dependencies>
 
   <build>
--- a/src/main/java/dev/morling/kccli/command/SetContextCommand.java
+++ b/src/main/java/dev/morling/kccli/command/SetContextCommand.java
@@ -15,20 +15,32 @@
  */
 package dev.morling.kccli.command;
 
+import java.net.URI;
+
+import dev.morling.kccli.service.Context;
 import dev.morling.kccli.util.ConfigurationContext;
 import picocli.CommandLine.Command;
 import picocli.CommandLine.Option;
+import picocli.CommandLine.Parameters;
 
 @Command(name = "set-context", description = "Set the context to cluster provided in arguments")
 public class SetContextCommand implements Runnable {
+    @Parameters(index = "0", description = "Context name")
+    String contextName;
 
-    @Option(names = { "--cluster" }, description = "URL of the Kafka Connect cluster to connect to", required = true)
+    @Option(names = { "--cluster" }, defaultValue = "http://localhost:8083", description = "URL of the Kafka Connect cluster to connect to")
     String cluster;
+
+    @Option(names = { "--bootstrap-servers" }, defaultValue = "localhost:9092", description = "Comma-separated list of Kafka broker URLs")
+    String bootstrapServers;
+
+    @Option(names = { "--offset-topic" }, description = "Name of the offset topic")
+    String offsetTopic;
 
     @Override
     public void run() {
         ConfigurationContext context = new ConfigurationContext();
-        context.setConfiguration(cluster);
-        System.out.println("Successfully set context to " + cluster);
+        context.setContext(contextName, new Context(URI.create(cluster), bootstrapServers, offsetTopic));
+        System.out.println("Using context " + contextName);
     }
 }
--- a/src/main/java/dev/morling/kccli/service/Context.java
+++ b/src/main/java/dev/morling/kccli/service/Context.java
@@ -1,0 +1,51 @@
+/*
+ *  Copyright 2021 The original authors
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package dev.morling.kccli.service;
+
+import java.net.URI;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+public class Context {
+    private final URI cluster;
+    private final String bootstrapServers;
+    private final String offsetTopic;
+
+    public Context(
+                   @JsonProperty("cluster") URI cluster,
+                   @JsonProperty("bootstrapServers") String bootstrapServers,
+                   @JsonProperty("offsetTopic") String offsetTopic) {
+        this.cluster = cluster;
+        this.bootstrapServers = bootstrapServers;
+        this.offsetTopic = offsetTopic;
+    }
+
+    public URI getCluster() {
+        return cluster;
+    }
+
+    public String getBootstrapServers() {
+        return bootstrapServers;
+    }
+
+    public String getOffsetTopic() {
+        return offsetTopic;
+    }
+
+    public static Context defaultContext() {
+        return new Context(URI.create("http://localhost:8083"), null, null);
+    }
+}
--- a/src/main/java/dev/morling/kccli/util/ConfigurationContext.java
+++ b/src/main/java/dev/morling/kccli/util/ConfigurationContext.java
@@ -15,41 +15,82 @@
  */
 package dev.morling.kccli.util;
 
+import java.io.File;
 import java.io.IOException;
-import java.net.URI;
-import java.nio.file.Files;
-import java.nio.file.Path;
-import java.nio.file.StandardOpenOption;
 
 import javax.enterprise.context.ApplicationScoped;
 
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.SerializationFeature;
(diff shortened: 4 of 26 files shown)


pom.xml
```diff
@@ -97,6 +97,16 @@
       <artifactId>quarkus-junit5</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.assertj</groupId>
+      <artifactId>assertj-core</artifactId>
```
gunnarmorling, line 102: Should be test scoped?

src/main/java/dev/morling/kccli/command/SetContextCommand.java
```diff
@@ -15,20 +15,32 @@
...
+
+import java.net.URI;
 
 @Command(name = "set-context", description = "Set the context to cluster provided in arguments")
 public class SetContextCommand implements Runnable {
+    @Parameters(index = "0", description = "Context name")
+    String contextName;
 
-    @Option(names = { "--cluster" }, description = "URL of the Kafka Connect cluster to connect to", required = true)
+    @Option(names = {"--cluster"}, description = "URL of the Kafka Connect cluster to connect to", required = true)
```
gunnarmorling, line 31: Should we default the options? http://localhost:8083 and localhost:9092 seem reasonable for KC and broker URLs? For the offset topic, we could leave that optional. If if it's not set for a given context, the future commands interacting with offsets wouldn't work without it, stating the need to specify a value in the chosen context. WDYT?

src/main/java/dev/morling/kccli/util/ConfigurationContext.java
```diff
@@ -15,40 +15,69 @@
...
+            objectMapper.writeValue(
+                configPath,
+                configuration.addConfigurationContext(contextName, context)
+            );
+        } catch (IOException e) {
+            throw new RuntimeException("Couldn't write context file ~/" + CONFIG_FILE, e);
         }
     }
 
     public URI getCluster() {
```
gunnarmorling, line 67: Should this method go away entirely, and all access be via `getContext()`? And we'd log an error if no context has been defined.

src/test/java/dev/morling/kccli/util/ConfigurationContextTest.java
```diff
@@ -0,0 +1,114 @@
...
+class ConfigurationContextTest {
+    @TempDir
+    File tempDir;
+
+    @Nested
+    class SetConfiguration {
+        @MethodSource("dev.morling.kccli.util.ConfigurationContextTest#setConfigurationArguments")
+        @ParameterizedTest
+        void should_create_a_new_configuration_when_no_configuration_exists(String contextName, Context context, String expectedConfiguration) throws IOException {
+            System.setProperty("user.home", tempDir.getAbsolutePath());
```
gunnarmorling, line 48: Neat trick :) Alternatively, there could be a constructor in `ConfigurationContext` which takes the config directory, while the default constructor would use the home dir.

src/main/java/dev/morling/kccli/util/ConfigurationContext.java
```diff
@@ -15,40 +15,69 @@
...
-            String config = Files.readString(CONFIG_PATH);
-            String[] parts = config.split("=");
-            return URI.create(parts[1]);
-        }
-        catch (IOException e) {
+            var configuration = objectMapper.readValue(configPath, Configuration.class);
+
+            return configuration.configurationContexts().get(configuration.getCurrentContext());
+        } catch (IOException e) {
             throw new RuntimeException("Couldn't read configuration file ~/" + CONFIG_FILE, e);
```
gunnarmorling, line 81: Let's add a sentence like "Create/update your configuration by running 'kcctl set-context <name> ...'".

src/main/java/dev/morling/kccli/service/Context.java
```diff
@@ -0,0 +1,48 @@
...
+ */
+package dev.morling.kccli.service;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import java.net.URI;
+
+public class Context {
+    private final URI cluster;
+    private final String bootstrapServers;
```
gunnarmorling, line 24: Should it be a list of strings (not sure from the top of my head about what the Kafka client expects)?

gunnarmorling (commented):
Nice, this LGTM overall. A few comments inline. Thx, @helpermethod!

gunnarmorling:
> One thing I would like to challenge is supporting the old, properties-like config format. Because we are still pre 1.0.0, I guess it would be acceptable to not support the current (properties-like) configuration format anymore. Wdyt?

Yes, that's ok. We only should log a meaningful error in that case.

Other than that, could you

* Rebase this PR so to address the current conflicts
* Squash the commits into one and prefix the message with a suitable emoj :)
* Add a section to the README describing the notion of contexts; two or three sentences will surely suffice, just to explain the overall approach

Thanks again!

src/test/java/dev/morling/kccli/util/ConfigurationContextTest.java
helpermethod: Had the exact same idea, but didn't want to change too much code at first. Good suggestion!

src/main/java/dev/morling/kccli/service/Context.java
helpermethod: Consumer and Producer Configs expect this as a single, comma-separated string 😢

src/main/java/dev/morling/kccli/util/ConfigurationContext.java
helpermethod: By logging you mean logging to STDOUT?

src/test/java/dev/morling/kccli/util/ConfigurationContextTest.java
helpermethod: Done.

src/main/java/dev/morling/kccli/command/SetContextCommand.java
helpermethod: Done.

pom.xml
helpermethod: Yep. Done.

src/main/java/dev/morling/kccli/util/ConfigurationContext.java
gunnarmorling: Yes, exactly.

src/main/java/dev/morling/kccli/service/Context.java
gunnarmorling: Ok, gotcha. Let's leave it as is then.

pom.xml
gunnarmorling: Are you still planning to push a commit here?

helpermethod: Incoming 🚀

helpermethod:
Now adding a few lines to the README.md.

helpermethod:
Hi @gunnarmorling, that push should resolve most, if not all of your comments 😄.

gunnarmorling:
@helpermethod, could you also re-generate the completion script (see instructions in the README) for the new options?

gunnarmorling:
Thanks a lot, @helpermethod! So I've played around a bit, it looks great overall. I wonder though if we can make the experience a bit smoother by default, i.e. if the user has not configured any context yet. How about the following if no context has been configured:

* When invoking any command other than `config` and its sub-commands, use http://localhost:8083 (as before), and display a warning such as "No configuration context has been defined, using http://localhost:8083 by default. Run ... to create a context."
* When invoking `config get-context`, print a message like "No configuration context has been defined, http://localhost:8083  will be used by default. Run ... to configure a create."

WDYT?

helpermethod:
Great suggestions, I'll incorporate them over the weekend!

gunnarmorling:
@helpermethod, if you'd like me to wrap up this one, just let me know. Would love to have it in, so we can get to the offset reset stuff :)

helpermethod:
Hi @gunnarmorling,

I'm on it, I'm currently stuffed with work so it's not easy to find some uninterrupted time 😄.

I don't quite get the difference between


> * When invoking any command other than config and its sub-commands, use http://localhost:8083 (as before), and display a warning such as "No configuration context has been defined, using http://localhost:8083 by default. Run ... to create a context."
> * When invoking config get-context, print a message like "No configuration context has been defined, http://localhost:8083 will be used by default. Run ... to configure a create."

So in any case basically a warning is printed and a default is returned?

gunnarmorling:
> So in any case basically a warning is printed and a default is returned?

Yes, that'd be the idea. Only the wording of the messages might be slightly different ("Using..." for actual commands vs. "Will use..." for get-context), but perhaps I'm overthinking it and we can have one which fits for both :)

gunnarmorling:
@helpermethod, there's a few odd changes part of this now, like version downgrades in the POM and removal of the Dependabot config.

helpermethod:
Messed up the latest commit, fixing it right now.

helpermethod:
I hope that should do it 😅!

gunnarmorling:
Yepp, LGTM. Applied. Thanks a lot!

PR #67, opened 2021-09-03
Fixed references to "kcctl set-context"

tonyfosterdev:
set-context is a sub-command of kcctl config.


gunnarmorling:
Good catch, thanks a lot! The context mechanism has changed a bit lately, and this seems to have been a small left-over.

Issue #61, opened 2021-08-23
Suppress stack trace in case of invalid configuration


gunnarmorling:
When trying to register a connector whose configuration is invalid, let's only display the actual validation error message(s), but not the stacktrace.

SiddiqueAhmad:
@gunnarmorling please review

gunnarmorling:
Fixed.

PR #74, opened 2021-09-27
handled exception for invalid connector config

SiddiqueAhmad:
- #61 handled exception and showed message rather stacktrace


gunnarmorling:
Squashed, added an emoji, and applied. Thx, @SiddiqueAhmad!

Issue #19, opened 2021-07-24
Allow to switch between contexts


gunnarmorling:
Having defined several named config context, it should be possible to show the current one and also switch:

```
$ kcctl config
Using configuration "local" (http://localhost:8083).

$ kcctl config use pre-prod
Using configuration "pre-prod" (http://qe:8083).

$ kcctl config use whoops
No configuration named "whoops" exists. Please run ...
```

gunnarmorling:
// CC @helpermethod for awareness.

helpermethod:
@gunnarmorling This requires a new field in `.kcctl` which contains the name of the current context. Maybe `current-context`?

SiddiqueAhmad:
we should use the same commands style as here 

kcctl config get-contexts                          # display list of contexts 
kcctl config current-context                       # display the current-context
kcctl config use-context my-cluster-name           # set the default context to my-cluster-name
kcctl config set-context --current           # to set the context and also mark it as current one

what you say @gunnarmorling ?

gunnarmorling:
+1, on both accounts, @helpermethod, @SiddiqueAhmad.

gunnarmorling:
There's now PR #48 by @helpermethod which puts the general concept scheme in place (essentially implementing `kcctl config set-context --current`). Once that's merged, we can make progress here along the lines described by @SiddiqueAhmad above.

gunnarmorling:
The basic set-context implementation is in place now.

SiddiqueAhmad:
#75 is sent, it is WIP, but can be merged so far

PR #75, opened 2021-09-27
#19 get-contexts, current-context, and get-context by name rather default fixed

SiddiqueAhmad:
- get-contexts command added
- current-context command added
- get-context command message changed
- get-context command removed default

### DIFF
--- a/src/main/java/dev/morling/kccli/command/CurrentContextCommand.java
+++ b/src/main/java/dev/morling/kccli/command/CurrentContextCommand.java
@@ -15,10 +15,22 @@
  */
 package dev.morling.kccli.command;
 
+import javax.inject.Inject;
+
+import dev.morling.kccli.util.ConfigurationContext;
+import picocli.CommandLine;
 import picocli.CommandLine.Command;
 
-@Command(name = "config", subcommands = { SetContextCommand.class, GetContextCommand.class, GetContextsCommand.class, CurrentContextCommand.class }, description = "Sets or retrieves the configuration of this client"
+@Command(name = "current-context", description = "Get the current context")
+public class CurrentContextCommand implements Runnable {
 
-)
-public class ConfigCommand {
+    @Inject
+    ConfigurationContext context;
+
+    @Override
+    public void run() {
+
+        String clusterUri = context.getCurrentContext().getCluster().toASCIIString();
+        System.out.println("Current context '"+context.getCurrentContextName()+"' is set to " + clusterUri);
+    }
 }
--- a/src/main/java/dev/morling/kccli/command/GetContextCommand.java
+++ b/src/main/java/dev/morling/kccli/command/GetContextCommand.java
@@ -18,17 +18,22 @@ package dev.morling.kccli.command;
 import javax.inject.Inject;
 
 import dev.morling.kccli.util.ConfigurationContext;
+import picocli.CommandLine;
 import picocli.CommandLine.Command;
 
 @Command(name = "get-context", description = "Get the current context")
 public class GetContextCommand implements Runnable {
+
+    @CommandLine.Parameters(index = "0", description = "Context name")
+    String contextName;
 
     @Inject
     ConfigurationContext context;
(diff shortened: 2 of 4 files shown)


src/main/java/dev/morling/kccli/command/CurrentContextCommand.java
```diff
@@ -0,0 +1,35 @@
...
+public class CurrentContextCommand implements Runnable {
+
+    @Inject
+    ConfigurationContext context;
+
+    @Override
+    public void run() {
+
+        String clusterUri = context.getCurrentContext().getCluster().toASCIIString();
+        System.out.println("Current context '" + context.getCurrentContextName() + "' is set to " + clusterUri);
```
gunnarmorling, line 33: Let's display all the info available from the context object, i.e. also bootstrap server, offset topic, etc. Ideally formatted using the tuple list approach, as for other outputs.

SiddiqueAhmad:
@gunnarmorling please review as I am going in right direction, I have added one test case too and modified one to test, current context and get context

src/main/java/dev/morling/kccli/command/GetContextsCommand.java
```diff
@@ -0,0 +1,40 @@
...
+
+    @Inject
+    ConfigurationContext configContext;
+
+    @Override
+    public void run() {
+        Map<String, Context> contexts = configContext.getContexts();
+        contexts.forEach((name, context) -> {
+            String clusterUri = context.getCluster().toASCIIString();
+            System.out.println(name + " is set to " + clusterUri);
```
gunnarmorling, line 37: Let's print this in table form here, with `NAME` and `URI`  as the headers.

src/main/java/dev/morling/kccli/command/GetContextCommand.java
```diff
@@ -18,17 +18,22 @@
...
+
     @Inject
     ConfigurationContext context;
 
     @Override
     public void run() {
-        String clusterUri = context.getContext().getCluster().toASCIIString();
-        System.out.println("Current context is set to " + clusterUri);
+
+        String clusterUri = context.getContext(contextName).getCluster().toASCIIString();
```
gunnarmorling, line 36: So this changes semantics of this command a bit, as it always has shown the currently active context. Can we keep it that way, when invoking `kcctl config get-context` without any option? And then, when calling `kcctl config get-context --name <NAME>`, show the specified one? I.e. make the name an optional parameter.

gunnarmorling (commented):
LGTM overall; some minor suggestions inline. Thx a lot, @SiddiqueAhmad!

src/main/java/dev/morling/kccli/command/GetContextCommand.java
SiddiqueAhmad: sure, that makes sense, I even did but could not get how to get the optional params, but your suggestion will help now to do it

src/main/java/dev/morling/kccli/command/GetContextsCommand.java
SiddiqueAhmad: sure, will do that

src/main/java/dev/morling/kccli/command/CurrentContextCommand.java
SiddiqueAhmad: okay I will do it

gunnarmorling:
@SiddiqueAhmad, you haven't pushed any update since the last round of feedback above, right?

SiddiqueAhmad:
> @SiddiqueAhmad, you haven't pushed any update since the last round of feedback above, right?

yes I am so sorry couldn't manage it yet, weekend was with debezium server

gunnarmorling:
No worries, just was curious. Didn't know you're looking into Debezium
Server too :)

>

SiddiqueAhmad:
sure, can't stop awesomeness to happen :)

PR #79, opened 2021-10-07
Kcctl 17 - Reforked

tonyfosterdev:
This PR can supercede https://github.com/kcctl/kcctl/pull/77.

@gunnarmorling Sorry for the mess with github here. I've added this PR as I didn't have permission to push to the branch kcctl-17 in the kcctl repo.  I reforked and pushed my changes up.

I added a global exception handler. My only question: should I let unexpected exceptions flow through so we can see the errors on the screen? Right now, I'm hiding them with a generic error message, and being more particular about 401 errors.

### DIFF
--- a/kcctl_completion
+++ b/kcctl_completion
@@ -401,7 +401,7 @@ function _picocli_kcctl_config_setcontext() {
 
   local commands=""
   local flag_opts=""
-  local arg_opts="--cluster --bootstrap-servers --offset-topic"
+  local arg_opts="--cluster --bootstrap-servers --offset-topic --username --password"
 
   compopt +o default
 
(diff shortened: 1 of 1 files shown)


gunnarmorling:
Excellent, thanks a lot!

> My only question: should I let unexpected exceptions flow through so we can see the errors on the screen? Right now, I'm hiding them with a generic error message, and being more particular about 401 errors.

Yeah, indeed I think we should let any other unexpected exception show up on the screen for now, as it's otherwise really hard to act on it. Upon future exception reports we'd then have to decide one by one on how to handle them more gracefully. A bit further out, we also may envision to only show the message by default but then have some sort of "--verbose" flag which displays more info including the exception. No need for that for now though, I think.

Apart from that fix, could you also update the shell completion file so it supports the new config command options. The process for doing so is described in the README.

Thanks again, I think we're almost there!

tonyfosterdev:
@gunnarmorling I just tested it, it seems to work great!

PR #80, opened 2021-10-08
Fix a few infelicities in README.md

gsmet:


gunnarmorling:
Gotcha :) Thanks for checking in! Loving the dev exp of Quarkus with
PicoCLI overall. The one thing I'm missing (out of your control though) is
simple CI on Arm, so we could publish binaries for that platform too. The
joys of native binaries ;)

>

Issue #83, opened 2021-10-10
Handle attempt to delete non-existing connector gracefully


gunnarmorling:
Currently we print the entire exception, but only the error message should be shown, as this not an exceptional situation.

gunnarmorling:
The command should return 1 when trying to delete a non-existent connector (i.e. KC replied with 404) and print out the error message. So far, we've printed out all messages on stdout, so I think we should keep doing that.

PR #85, opened 2021-10-15
#83 added exception handler

SiddiqueAhmad:
added exception handler, but could not run test, context injection not working, any help is appreciated.﻿

### DIFF
--- a/src/main/java/org/moditect/kcctl/command/DeleteConnectorCommand.java
+++ b/src/main/java/org/moditect/kcctl/command/DeleteConnectorCommand.java
@@ -20,6 +20,7 @@ import javax.inject.Inject;
 import org.eclipse.microprofile.rest.client.RestClientBuilder;
 import org.moditect.kcctl.completion.ConnectorNameCompletions;
 import org.moditect.kcctl.service.KafkaConnectApi;
+import org.moditect.kcctl.service.KafkaConnectException;
 import org.moditect.kcctl.util.ConfigurationContext;
 
 import picocli.CommandLine.Command;
@@ -36,11 +37,15 @@ public class DeleteConnectorCommand implements Runnable {
 
     @Override
     public void run() {
+
         KafkaConnectApi kafkaConnectApi = RestClientBuilder.newBuilder()
                 .baseUri(context.getCurrentContext().getCluster())
                 .build(KafkaConnectApi.class);
-
-        kafkaConnectApi.deleteConnector(name);
-        System.out.println("Deleted connector " + name);
+        try {
+            kafkaConnectApi.deleteConnector(name);
+            System.out.println("Deleted connector " + name);
+        } catch (KafkaConnectException kafkaConnectException) {
+            System.out.println(kafkaConnectException.getMessage());
+        }
     }
 }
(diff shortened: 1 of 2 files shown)


SiddiqueAhmad:
> Perhaps just instantiate the subcommand yourself? E.g. like so:
> 
> ```
> DeleteConnectorCommand cmd = new DeleteConnectorCommand();
> cdm.ctx = new ConfigurationContext();
> ...
> ```

I tried, but ctx not available, I am seeing injection working, but it is being lost while it is running command,

tonyfosterdev:
@SiddiqueAhmad If you look at the ConfigurationContextTests, you can see what is done there.

Basically, a .kcctl file is generated in the test and stored in a temp directory. It drives the test.

I'd link it but I'm on my phone, give that a shot!

tonyfosterdev:
@SiddiqueAhmad look here's this map help: https://github.com/kcctl/kcctl/blob/f27170156cbcd45c82cb8e540cd648fa76173b58/src/test/java/org/moditect/kcctl/util/ConfigurationContextTest.java#L76

tonyfosterdev:
Ignore my previous notes, I see what you were talking about. To Gunnar's note above, did you try:

```java
DeleteConnectorCommand command = new DeleteConnectorCommand();
command.context = new ConfigurationContext();
command.run();
```

Doing that instead of calling through KcCtlCommand may be easier.

SiddiqueAhmad:
@tonyfosterdev thanks for giving time to this, there is no context available for command object

tonyfosterdev:
I believe this is a good place to make  context a property with a getter and setter. The getter would default to the default constructed ConfigurationContext unless set otherwise.

If we like this, it would be a good pattern to formalize via interface or abstraction.

Any objections to this approach?

SiddiqueAhmad:
@gunnarmorling can comment better,

gunnarmorling:
@tonyfosterdev, I don't quite follow what you suggest exactly. Note that ideally all the fields of any class would be `private final`, i.e. immutable. It's not the case, as it simplifies dependency injection for Quarkus. But the fields already can be set manually for tests (as shown above), so I'm not quite sure what adding a setter would gain? Can you expand a bit?

gunnarmorling:
Ok, gotcha (and yeah, envying you a tiny bit for the property approach in C# ;); so as you say, ideally, we'd use constructor injection as well. I'm just not sure whether it is supported by Quarkus; it definitely _is_ supported by CDI, the underlying DI spec defining `@Inject` etc. (Quarkus doesn't implement it 100% though). The following actually may work:

```
class DeleteConnectorCommand {

  private final ConfigurationContext configurationContext;

  @Inject
  public DeleteConnectorCommand(ConfigurationContext configurationContext) {
    this.configurationContext = configurationContext;
  }
}

And in the test:

```
// Arrange
ConfigurationContext myTestContext = new ConfigurationContext('test-kcctl-file');
DeleteConnectorCommand command = new DeleteConnectorCommand(myTextContext);

// Act
command.run();

// Assert something
```

I.e. in non-test usage, Quarkus should invoke the `@Inject`-annotated constructor and pass the context.

gunnarmorling:
@SiddiqueAhmad so what should we do about this one? Are you planning to wrap it up? Taking a step back, what actually was the problem we're trying to solve here?

SiddiqueAhmad:
thanks, sure I tried but I have to learn more java to solve this, 

I would love to conclude it

we were looking for configuration context in our command

gunnarmorling:
Ok, so I remembered now: this was about gracefully handling the attempt of deleting a non-existing connector. In fact, this has been addressed already via https://github.com/kcctl/kcctl/pull/111, hence I'm going to close this one. Thx, @SiddiqueAhmad!

PR #106, opened 2021-11-25
#105 Fix file paths to match package declarations

### LINKED ISSUE #105: File path should match the package declarations
Commit 5839388ba4c78d9dd6e2f9ea2958772dd2aef783  renamed `org.moditect` to `org.kcctl` throughout the whole project, but the top level folder wasn't updated accordingly. Maven build passes however the IDEs (in my case IntelliJ) would complain and fail to compile with this error: `Package name 'org.kcctl.command' does not correspond to the file path 'org.moditect.kcctl.command'`.

iabudiab:
Commit 5839388 renamed org.moditect to org.kcctl throughout the whole
project, but the top level folder wasn't updated accordingly. Maven
build passes however many IDEs, including IntelliJ, would complain and
fail to compile with the following error:
"Package name 'org.kcctl.command' does not correspond to the file
path 'org.moditect.kcctl.command'."

Fixes #105


gunnarmorling:
Whoops, good spot!

gunnarmorling:
Applied, thx!

Issue #98, opened 2021-11-14
Output formats


tonyfosterdev:
It would be cool to enable various format options for output. We could call the current format "human", but json would be cool too.

Kubectl supports this with the -o flag.

The main use case here would be to allow for scripting support in conjunction with tools like jq.

gunnarmorling:
Great idea!

jornh:
@tonyfosterdev at least for the `describe` command as a workaround you can use [`yq`](https://mikefarah.gitbook.io/yq/usage/convert) to convert to json - as the format you call "human" is actually `yaml`.

There is _one caveat_ though. The yaml has color codes in some places which yq chokes on. To get past that we need another workaround and use `sed to strip that out. The end result:

``` bash
kcctl describe connector http3 | sed 's/\x1b\[[0-9;]*m//g' | yq eval -o json
```

gunnarmorling:
That's an interesting observation, @jornh. Note that the YAML output for describe is mostly incidental though, it was not a conscious decision and we may derive from this being YAML in some cases.

juanlucruz:
I am lacking this issue exactly. I've tried with the workaround using yq but it's not working for me. Thanks anyway for the wonderful tool Gunnar!

Issue #44, opened 2021-08-14
Have tests for kcctl


gunnarmorling:
Currently, no tests exist for kcctl at all. We should add tests for the different commands, so to avoid regressions down the road. As part of this task, we need to

* Understand how PicoCLI applications are tested; is their any guidance in their docs?
* Set up the infrastructure for integration tests via Testcontainers; we'll need to start up ZooKeeper, Kafka, and Kafka Connect for testing kcctl; as inspiration or starting point, here's a [Docker Compose](https://github.com/gunnarmorling/kcetcd/blob/main/docker-compose.yaml) file which shows the required components; as a test connector, the built-in file connector could probably be used

SiddiqueAhmad:
I setup the container resources class took help from debezium facing few issues, but we will do it

karim-elngr:
The picocli documentation provide some helpful insights into [writing unit tests](https://picocli.info/#_testing_your_application)

gunnarmorling:
Thanks, @karim-elngr, so essentially we'd just do assertions on the output stream we've set before. Makes sense. We can throw AssertJ into the mix for more expressive assertions.

@SiddiqueAhmad, ok, excellent. Seems the way forward to having some first integration tests is pretty clear then.

karim-elngr:
I think this is worth mentioning that SDK man uses a very interesting BDD approach for testing, I wonder if there is a common approach to test cli tools

https://github.com/sdkman/sdkman-cli/blob/master/src/test/resources/features/java_installation.feature

gunnarmorling:
That's done now thanks to a heroic effort by @iabudiab 🎉 .

Issue #32, opened 2021-08-13
Gracefully handle attempt to instantiate non-existent connector type


gunnarmorling:
Instead of showing the raw REST response, display the properly formatted list of available connector plug-ins.

gunnarmorling:
Done.

PR #110, opened 2021-11-28
[WIP] #44 Integration tests setup

iabudiab:
Hey there 👋 I am working on a setup for integration tests with testcontainers. The first test runs w/o problems (on my machine 😁) 

However, this is a WIP right now because I wanted to ping-pong some details here:

- One way to test the commands is to assert the return codes and their outputs. Using the `@Spec CommandSpec` from `picocli` allows for easier testing. See the changes in the `InfoCommand` in this PR. Basically a `CommandSpec` can be injected vie the `@Spec` annotation and then used for writing the output. In the test we can then set the output stream of the `CommandLine` like this and then assert on it:

```java
CommandLine commandLine = new CommandLine(infoCommand);
StringWriter output = new StringWriter();
commandLine.setOut(new PrintWriter(output));
```

- A better way would be to assert on the JSON (#98) instead of the "human" format. So maybe I could focus the efforts there before finishing this one.

- The commands should use constructor injection instead of field injection for them to be testable. i.e.

```java
@Inject
public InfoCommand(ConfigurationContext context) {
     this.context = context;
}
```

- Better yet, the whole `RestClient` should be somehow injectable. That way we could write unit and integration tests. Right now only blackbox testing is easily doable (I think), because the client is initialised when the command runs. Maybe someone with more quarks experience could chime in here 😉 

- Do we have some guidelines/conventions for the status codes and the output streams? i.e. out vs err? I am asking, because for example what would be the better approach for #83? Status 0 and stdout? or some non-zero status with stderr? or status 0 but stderr?

- I've used a `@Tag` for the integration test, in order to be able to run them selectively:

```shell
# run all tests
mvn test

# run only integration tests
mvn test -Dgroups=integration-test

# run all except integration tests
mvn test -Dgroups=!integration-test
```

What do you think?

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -109,6 +109,23 @@
       <scope>test</scope>
     </dependency>
 
+    <dependency>
+      <groupId>io.debezium</groupId>
+      <artifactId>debezium-core</artifactId>
+      <version>1.6.1.Final</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>io.debezium</groupId>
+      <artifactId>debezium-testing-testcontainers</artifactId>
+      <version>1.6.1.Final</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.testcontainers</groupId>
+      <artifactId>kafka</artifactId>
+      <scope>test</scope>
+    </dependency>
   </dependencies>
 
   <build>
--- a/pom.xml
+++ b/pom.xml
@@ -28,8 +28,8 @@
     <compiler-plugin.version>3.8.1</compiler-plugin.version>
     <executable-suffix/>
     <maven.compiler.parameters>true</maven.compiler.parameters>
-    <maven.compiler.source>11</maven.compiler.source>
-    <maven.compiler.target>11</maven.compiler.target>
+    <maven.compiler.source>17</maven.compiler.source>
+    <maven.compiler.target>17</maven.compiler.target>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
     <quarkus.platform.artifact-id>quarkus-universe-bom</quarkus.platform.artifact-id>
@@ -112,18 +112,37 @@
     <dependency>
       <groupId>io.debezium</groupId>
       <artifactId>debezium-core</artifactId>
-      <version>1.6.1.Final</version>
+      <version>1.8.0.Final</version>
+      <type>test-jar</type>
       <scope>test</scope>
     </dependency>
     <dependency>
       <groupId>io.debezium</groupId>
       <artifactId>debezium-testing-testcontainers</artifactId>
-      <version>1.6.1.Final</version>
+      <version>1.8.0.Final</version>
       <scope>test</scope>
     </dependency>
     <dependency>
       <groupId>org.testcontainers</groupId>
       <artifactId>kafka</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>com.squareup.okhttp3</groupId>
+      <artifactId>okhttp</artifactId>
+      <version>4.9.1</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>com.squareup.okhttp3</groupId>
+      <artifactId>okhttp-urlconnection</artifactId>
+      <version>4.9.1</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>com.squareup.okio</groupId>
+      <artifactId>okio</artifactId>
+      <version>2.8.0</version>
       <scope>test</scope>
     </dependency>
   </dependencies>
@@ -161,6 +180,7 @@
             <groups>java.,javax.,org.,com.</groups>
             <removeUnused>true</removeUnused>
             <staticAfter>true</staticAfter>
+            <compliance>17</compliance>
             <!-- <staticGroups>java.,javax.,org.w3c.,org.xml.,junit.</staticGroups> -->
           </configuration>
         </plugin>
--- a/src/main/java/org/kcctl/command/InfoCommand.java
+++ b/src/main/java/org/kcctl/command/InfoCommand.java
@@ -23,12 +23,21 @@ import org.kcctl.service.KafkaConnectInfo;
 import org.kcctl.util.ConfigurationContext;
 
 import picocli.CommandLine.Command;
+import picocli.CommandLine.Model.CommandSpec;
+import picocli.CommandLine.Spec;
 
 @Command(name = "info", description = "Displays information about the Kafka Connect cluster")
 public class InfoCommand implements Runnable {
 
+    private final ConfigurationContext context;
+
+    @Spec
+    CommandSpec spec;
+
     @Inject
-    ConfigurationContext context;
+    public InfoCommand(ConfigurationContext context) {
+        this.context = context;
+    }
 
     @Override
     public void run() {
@@ -37,9 +46,9 @@ public class InfoCommand implements Runnable {
                 .build(KafkaConnectApi.class);
 
         KafkaConnectInfo workerInfo = kafkaConnectApi.getWorkerInfo();
-        System.out.println("URL:               " + context.getCurrentContext().getCluster());
-        System.out.println("Version:           " + workerInfo.version);
-        System.out.println("Commit:            " + workerInfo.commit);
-        System.out.println("Kafka Cluster ID:  " + workerInfo.kafka_cluster_id);
+        spec.commandLine().getOut().println("URL:               " + context.getCurrentContext().getCluster());
+        spec.commandLine().getOut().println("Version:           " + workerInfo.version);
+        spec.commandLine().getOut().println("Commit:            " + workerInfo.commit);
+        spec.commandLine().getOut().println("Kafka Cluster ID:  " + workerInfo.kafka_cluster_id);
     }
 }
--- a/src/test/java/org/kcctl/IntegrationTest.java
+++ b/src/test/java/org/kcctl/IntegrationTest.java
@@ -15,24 +15,32 @@
  */
 package org.kcctl;
 
-import java.io.File;
 import java.io.IOException;
+import java.io.PrintWriter;
+import java.io.StringWriter;
+import java.lang.reflect.*;
 import java.nio.file.Files;
 import java.util.stream.Stream;
 
+import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.BeforeEach;
-import org.junit.jupiter.api.Tag;
-import org.junit.jupiter.api.io.TempDir;
+import org.junit.jupiter.api.extension.ExtensionConfigurationException;
+import org.junit.jupiter.api.extension.ParameterResolutionException;
+import org.junit.platform.commons.util.ReflectionUtils;
+import org.junit.platform.commons.util.ReflectionUtils.HierarchyTraversalMode;
+import org.kcctl.support.InjectCommandContext;
+import org.kcctl.support.KcctlCommandContext;
 import org.kcctl.util.ConfigurationContext;
 import org.testcontainers.containers.KafkaContainer;
 import org.testcontainers.containers.Network;
 import org.testcontainers.lifecycle.Startables;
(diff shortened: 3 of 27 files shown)


gunnarmorling:
> One way to test the commands is to assert the return codes and their outputs. Using the @Spec CommandSpec from picocli allows for easier testing. See the changes in the InfoCommand in this PR. Basically a CommandSpec can be injected vie the @Spec annotation and then used for writing the output. In the test we can then set the output stream of the CommandLine like this and then assert on it

Ah, that's cool, didn't know about this. Seems like the way to go; in fact, we already have some tests where we redirect the output stream, ideally we'd rework those then to use the spec approach.

> A better way would be to assert on the JSON

You mean better as it'd be easier to parse the output and retrieve specific bits of information? If so, that sounds good too. OTOH testing the human-readable output would also be valuable. Both would work for me.

In regards to (constructor) injection of dependencies, I'll be happy with everything which helps to add some tests. That said, I'm not a big believer into mocked tests, so I think my preference would always be integration tests which actually test against a running Kafka Connect. Note with offset display and reset (#2, #8), we'll also interact with Kafka proper, which would make mocking even more complex.

> Do we have some guidelines/conventions for the status codes and the output streams? i.e. out vs err?

Good question; no explicit guidelines so far. We should return a value other than 0 if there's an error (that should be the case already pretty consistently; where it isn't we should change it). In terms of stdout vs. stderr, I have no preference tbh. Will contents from stderr be visible by default on the console? I've only ever used stdout so far. Replied on #32, too.

> I've used a @Tag for the integration test, in order to be able to run them selectively

Ah yes, that's very nice.

Thanks a lot for driving this forward, @iabudiab, and welcome to the project!

src/main/java/org/kcctl/command/InfoCommand.java
```diff
@@ -23,12 +23,21 @@
...
+import picocli.CommandLine.Model.CommandSpec;
+import picocli.CommandLine.Spec;
 
 @Command(name = "info", description = "Displays information about the Kafka Connect cluster")
 public class InfoCommand implements Runnable {
 
+    private final ConfigurationContext context;
+
+    @Spec
+    CommandSpec spec;
```
gunnarmorling, line 35: So can we get this via the constructor too? Would be nice if we could make all the fields `private final`.

src/test/java/org/kcctl/IntegrationTest.java
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    @BeforeEach
+    void setupContext(@TempDir File tempDir) throws IOException {
+        var configFile = tempDir.toPath().resolve(".kcctl");
+
+        Files.writeString(configFile, String.format(
+                "{ \"currentContext\": \"local\", \"local\": { \"cluster\": \"%s\", \"username\": \"testuser\", \"password\": \"testpassword\" }}",
+                kafkaConnect.getTarget())
+        );
```
gunnarmorling, line 62: So the latest Quarkus and GraalVM versions should work with Java 17, i.e. we could mandate that version for building kcctl and take advantage of multi-line strings here. Given that kcctl is distributed as self-contained native binaries anyways, there's no concern really about ensuring backwards compatibility as with JVM based distributables.

gunnarmorling (approved):
This looks great overall; a few comments inline. Thanks a lot, @iabudiab!

pom.xml
```diff
@@ -109,6 +109,23 @@
       <scope>test</scope>
     </dependency>
 
+    <dependency>
+      <groupId>io.debezium</groupId>
+      <artifactId>debezium-core</artifactId>
+      <version>1.6.1.Final</version>
+      <scope>test</scope>
+    </dependency>
```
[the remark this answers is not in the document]
iabudiab: I'm not 100% sure. If I remember correctly it is/was required for some version. I'll check and remove accordingly if it's not the case.

src/main/java/org/kcctl/command/InfoCommand.java
iabudiab: Not easily, since then we would have to register picocli CommandSpecs for quarkus to somehow be injectable. I could take a deeper look into that.

src/test/java/org/kcctl/IntegrationTest.java
iabudiab: 👍 I'm for it. I'll be building and testing with Java 17 from now on.

iabudiab:
Thanks for the feedback and the warm welcome 😄  I'll pick up and continue this PR sometime in the next couple of days.

src/test/java/org/kcctl/IntegrationTest.java
gunnarmorling: Quick heads-up: I've increased the baseline for this project to Java 17 (see #113), trying to build with an earlier version will fail. I've also adjusted one test to use the text block syntax.

src/test/java/org/kcctl/IntegrationTest.java
```diff
@@ -0,0 +1,66 @@
...
+import org.kcctl.util.ConfigurationContext;
+import org.testcontainers.containers.KafkaContainer;
+import org.testcontainers.containers.Network;
+import org.testcontainers.lifecycle.Startables;
+import org.testcontainers.utility.DockerImageName;
+
+import io.debezium.testing.testcontainers.DebeziumContainer;
+
+@Tag("integration-test")
+public abstract class IntegrationTest {
```
Thomas-Dangleterre, line 36: Maybe you could implements [QuarkusTestResourceLifecycleManager](https://quarkus.io/guides/getting-started-testing#quarkus-test-resource) ?

gunnarmorling: +1, that's a great idea.

iabudiab:
👋 just wanted to check-in and let you all know, I am still working on this, however the progress is really slow because I am on (long overdue) vacation now 😸 
I won't be able to make real progress until after Christmas 🎄☃️
Cheers and happy holidays everybody

gunnarmorling:
Hey @iabudiab, thanks for the heads-up. No worries, looking forward to this one, once you find the time for it. Have a great vacation, and Happy Holidays to you, too!

iabudiab:
Happy new year everybody!

So, here is a summary of the current state of this PR and some of the gotchas I collected while on a trip down the rabbit hole :) _(This is my first time using quarkus)_

Stuff good to know to reduce the time spent (re)searching (in no particular order):
- Quarkus and JUnit5 integration is not optimal. For example you can't use JUnit5 extensions with `@QuarkusTest`s because quarkus uses a separate class loader when instantiating a test instance => The test instance in the extension is not the same that is actually used for the test, so that almost everything you do in the extension is not seen in the test.
- Using `io.debezium:debezium-testing-testcontainers` version > 1.6.x requires `io.debezium:debezium-core:test-jar` because of this one class `io.debezium.util.ContainerImageVersions`. It should either bring the dependency transitively or this class should be refactored somewhere else maybe? @gunnarmorling  maybe you can sort it out on the Debezium side?
- `quarkus-universe-bom` pins `com.squareup.okhttp3:okhttp` to version `3.14.9`, which is fine, unless you are using `DebeziumContainer` in order to create connectors in the running container. `io.debezium:debezium-testing-testcontainers` requires at least `okhttp:4.x`. So without `okhttp:4.x` dependency you would get `NoSuchMethodError`s and `NoSuchFieldError` when using certain methods from `DebeziumContainer`, e.g. registering a connector.
- Quarkus test callbacks (i.e. `AfterTestCallback`, `BeforeEachTestCallback` etc.) can only be deployed as service provider on the class path, which would activate them for all the tests. There is currently no way (from what I've seen) to have test callbacks analogous to JUnit5 extension, i.e. activatable via annotations, when using `@QuarkusTest`s.

This is why I chose to implement the simplest options because of the limited time that I have, i.e. an abstract super class for integration tests, that contains all the required utilities and where all subclasses can access stuff where needed.

However, I pushed the other two options, that I've tried onto separate branches, in case somebody with more quarkus experience than me wants to pick those up and continue from there or give some pointers 😉. You can find those [here: quarkus_test_resource](https://github.com/iabudiab/kcctl/tree/%2344_quarkus_test_resource) and [here: junit_extension](https://github.com/iabudiab/kcctl/tree/%2344_junit_extension). Take a look at the test class `org.kcctl.command.ApplyComamnd` and related code.

I've implemented some integration tests now for several commands and will push the rest whenever I get the time.

Cheers.

src/test/java/org/kcctl/IntegrationTest.java
gunnarmorling: Any thoughts on this one, @iabudiab?

src/test/java/org/kcctl/IntegrationTest.java
```diff
@@ -0,0 +1,159 @@
...
+import picocli.CommandLine;
+
+public abstract class IntegrationTest {
+
+    protected static final Network network = Network.newNetwork();
+
+    protected static final KafkaContainer kafka = new KafkaContainer(DockerImageName.parse("confluentinc/cp-kafka:5.4.3"))
+            .withNetwork(network);
+
+    protected static DebeziumContainer kafkaConnect = new DebeziumContainer(DockerImageName.parse("debezium/connect:1.7.1.Final"))
```
gunnarmorling, line 51: Let's use 1.8.0.Final.

gunnarmorling (approved):
Hey @iabudiab, some very small comments inline, but overall this looks *awesome*! Thank you so much for this excellent piece of work, it's a huge step forward for the project to have this test suite.

src/main/java/org/kcctl/command/InfoCommand.java
iabudiab: I've looked into this a little bit. However, I couldn't find an easy/direct way to make this work with Quarkus. It seems that supporting this in PicoCLI directly is the best option.

src/test/java/org/kcctl/IntegrationTest.java
iabudiab: Yup, will do ;)

src/test/java/org/kcctl/IntegrationTest.java
iabudiab: I've made a PoC implementation of this on a [separate branch](https://github.com/iabudiab/kcctl/tree/%2344_quarkus_test_resource) in case this is the way we want to go forward. However, I wasn't really happy with the way it integrates with JUnit and the glue it requires. And because of my limited time and Quarkus experience, I decided not pursue it further at this point. Of course, if we decide that this is nonetheless the better way, then we have that option too ;)

Please see my comment for more details: https://github.com/kcctl/kcctl/pull/110#issuecomment-1008474366

iabudiab:
@gunnarmorling I'm glad to help 😄 I wished I had more free time to move things along faster.

src/test/java/org/kcctl/IntegrationTest.java
gunnarmorling: Took care of it in https://github.com/kcctl/kcctl/pull/144

gunnarmorling:
>  It should either bring the dependency transitively or this class should be refactored somewhere else maybe? @gunnarmorling maybe you can sort it out on the Debezium side?

Took a look into this, and the dependency is already established correctly by _debezium-testing-testcontainers_. The problem is though that the Quarkus BOM brings in _debezium-core_ 1.6.1, whereas here we explicitly depend on _debezium-testing-testcontainers_ 1.8.0. So I have pinned the version for core to 1.8.0 via the dependency management within kcctl, and then things work as they should, without adding the transitive core dependency.

> quarkus-universe-bom pins com.squareup.okhttp3:okhttp to version 3.14.9, which is fine, unless you are using DebeziumContainer in order to create connectors in the running container.

Ah, that's interesting. Thinking about it, I don't think we should use okhttp in Debezium to begin with, but rather use the HTTP client coming with Java 11+. I've logged https://issues.redhat.com/browse/DBZ-4594 for this.

Regarding the Quarkus integration issues you encountered, I think the base class approach is good enough. I'll log a follow-up issue, referencing the branches you linked to above, in case you or someone else will have the time to look into it. It'd be a bit more "idiomatic" and doing things the Quarkus way, but it doesn't really change anything in terms of functionality, and the tests you added provide excellent value as-is.

I have rebased everything to the latest main to resolve some merge conflicts and added a few fixes. All this is in PR https://github.com/kcctl/kcctl/pull/144, which supersedes this one. I'll close this and merge the other one in second. Thanks a lot, @iabudiab, this is really a fantastic contribution to kcctl, substantially improving the trust we can have into future changes, and helping to spot any regressions. Thank you!

PR #111, opened 2021-11-29
fix: Handle attempt to delete non-existing connector gracefully

### LINKED ISSUE #83: Handle attempt to delete non-existing connector gracefully

Thomas-Dangleterre:
Fixes #83


gunnarmorling:
Thanks a lot, applied.

Issue #116, opened 2021-12-01
Kcctl apply - Read from stdin 


Thomas-Dangleterre:
Hi,

It would be nice to be able to read the connector configuration from stdin when calling `kcctl apply`. 

As --file option is required, it might be handled like Kubernetes by reading stdin when the option is equals to "-" : 
> Apply the JSON passed into stdin to a pod
> 
> cat pod.json | kubectl apply -f -

cf the kubectl [doc](https://kubernetes.io/docs/reference/generated/kubectl/kubectl-commands#apply) :

gunnarmorling:
Interesting proposal; do you have a specific use case in mind for this? Some kind of dynamically generated config perhaps? In any case, seems like a reasonable idea to support reading config from stdin, alone for being a good CLI citizen. Wanna give it a try?

Thomas-Dangleterre:
Yes, our connectors' configurations are injected to our pods as environnement variables.

I will gladly provide a PR :)

gunnarmorling:
Excellent, thanks a lot and looking forward to the PR! Out of curiosity, couldn't you expose those connector configs as a config map alternatively? And as you are running on Kubernetes, have you looked into using [Strimzi](https://strimzi.io) for deploying Kafka, Connect, and connectors via custom resources? Not that I want to argue against the proposed change (I really like it), just curious about your usage :)

Thomas-Dangleterre:
We use Helm with the image :  confluentinc/cp-kafka-connect on GKS, we did this as we copied the way we deploy our apps. We also use Vault not only as a secret manager but also as a configuration manager ( thanks to banzai cloud ) so that's why our connectors' configurations are injected as env variables.

I will take a look at Strimzi, Thanks.

gunnarmorling:
And applied, great improvement! Thanks, @Thomas-Dangleterre!

PR #117, opened 2021-12-01
feat: kcctl apply - Read from stdin

Thomas-Dangleterre:
Implements #116


src/main/java/org/kcctl/command/ApplyCommand.java
```diff
@@ -44,7 +45,7 @@
     @Inject
     ConfigurationContext context;
 
-    @Option(names = { "-f", "--file" }, description = "Name of the file to apply", required = true)
+    @Option(names = { "-f", "--file" }, description = "Name of the file to apply or '-' to read from stdin", required = true)
```
gunnarmorling, line 48: Could you also update the description of the command itself (and copy over the output into the README, which currently serves as the reference for all commands).

src/main/java/org/kcctl/command/ApplyCommand.java
```diff
@@ -86,6 +91,14 @@ public Integer call() throws Exception {
         }
     }
 
+    private String readFromStdin() {
+        Scanner sc = new Scanner(System.in);
```
gunnarmorling, line 95: The scanner instance should be closed at the end?

gunnarmorling (commented):
LGTM, thanks a lot. Some minor comments inline. Too bad we don't have the testing infra in place yet (that's happening right now via a parallel PR).

gunnarmorling:
Applied, thanks a lot!

Issue #127, opened 2021-12-21
Deploy to Home Brew


gunnarmorling:

aalmiray:
Fixed with https://github.com/kcctl/kcctl/pull/128

hartmut-co-uk:
by the way I'm getting 

![image](https://user-images.githubusercontent.com/10864443/147226820-e98b89bd-28e5-4413-b790-b367641a6d03.png)

PR #128, opened 2021-12-21
🍺 Add Homebrew packager

aalmiray:
Fixes https://github.com/kcctl/kcctl/issues/127

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,12 @@ The latest stable release of _kcctl_ (x86) for Linux, macOS, and Windows can be 
 sdk install kcctl
 ```
 
+You may also use [Homebrew](https://brew.sh/) to install _kcctl_ on Linux and macOs, by configuring our tap
+
+```shell script
+brew install kcctl/tap/kcctl
+```
+
 It is recommended to install the bash/zsh completion script _kcctl_completion_:
 
 ```shell script
--- a/pom.xml
+++ b/pom.xml
@@ -412,6 +412,10 @@
                   </snapshot>
                 </project>
                 <packagers>
+                  <brew>
+                    <active>RELEASE</active>
+                    <multiPlatform>true</multiPlatform>
+                  </brew>
                   <sdkman>
                     <active>RELEASE</active>
                   </sdkman>


gunnarmorling:
Ok, but that can't be all, right? I suppose I need to request deploy permissions for brew as well?

aalmiray:
Oh but it is, because the formula is pushed to your own tap, not the main Homebrew repo.

gunnarmorling:
Ok, and that's validated against this org's name then? So no one else can publish to kcctl's identity?

gunnarmorling:
Applied in any case, thx! Seems I should do another release then?

gunnarmorling:
So reading https://docs.brew.sh/How-to-Create-and-Maintain-a-Tap, I don't fully get it yet. Do we need to create a _kcctl/homebrew-repo_ repo or something like that?

hartmut-co-uk:
I'm getting

image

is this me only?

helpermethod:
`So reading https://docs.brew.sh/How-to-Create-and-Maintain-a-Tap, I don't fully get it yet. Do we need to create a kcctl/homebrew-repo repo or something like that?`

JReleaser creates the repo automatically.

gunnarmorling:
@helpermethod, yes, that has all been sorted out by now; the repo has been created, and kcctl should be available via Homebrew (it works for me, but it seems @hartmut-co-uk is having a problem).

gunnarmorling:
@hartmut-co-uk, which OS and architecture are you on? AArch64 by any chance? If so, that'd explain; we currently publish binaries for x86 only.

hartmut-co-uk:
AArch64 yes

gunnarmorling:
Ok, I see. So we should raise a more meaningful error in the tap in that situation // CC @aalmiray.

We'll have to see whether/how we can publish binaries for AArch64/Linux (there's no easy-to-use runner for that for GH Actions). If you have GraalVM with the [native image](https://www.graalvm.org/reference-manual/native-image/) tool installed, you could build it yourself locally via `mvn clean verify -Pnative`.

hartmut-co-uk:
FYI I just downloaded the latest osx binaries (https://github.com/kcctl/kcctl/releases/tag/v1.0.0.Alpha5) and can run without problems (very basic first try)

gunnarmorling:
Ah, that's great to hear. So the x64 emulation on the M1 works for this as expected then, very nice. Didn't think of that initially.

In that light, we may publish the x64/macOS one for AArch64/macOS then, too. @aalmiray, WDYT?

hartmut-co-uk:
FYI: I've done some more testing / actually used kcctl today, everything works just fine!

aalmiray:
@gunnarmorling how will we build a Native Image for osx-aarch64? Based on the list of binaries listed at https://github.com/graalvm/graalvm-ce-builds/releases we can only build native executables for

- linux-x86_64 (matches linux-amd64)
- linux_aarch64
- osx-x86_64 (matches darwin-amd64)
- windows-x86_64 (matches windows-amd64)

We can't build for `linux-aarch64` using a default GitHub runner _unless_ we use a Docker image running inside an ubuntu-latest runner (yay containers!).

gunnarmorling:
> how will we build a Native Image for osx-aarch64?

IIUC, we could simply take the x86 binary and distribute that to macOS/AArch64 users too. macOS on the M1 comes with a powerful emulation layer which should allow to run that one as-is. I think it's only a question of configuring the right metadata in the Brew formula for taking the x86 binary also for AArch64.

aalmiray:
If that's the case then duplicating the existing macOS artifact definition found in the `pom.xml` but using `<platform>osx-aarch_64</platform>` would be enough. The name of the file remains the same but that's of no consequence to the brew formula.

gunnarmorling:
Logged https://github.com/kcctl/kcctl/issues/134 for this.

PR #167, opened 2022-03-04
Allow to manage multiple connectors when it apply

### LINKED ISSUE #107: Allow multiple connectors to be created/updated with kcctl apply
It should be possible to provide multiple files:

    kcctl apply --f my-connector1.json --f my-connector2.json

    Created connector my-connector1.
    Updated connector my-connector2.

An open question is how to deal with names if not provided within the files.

jvenant:
Fixes #107 

This is a base of discussion about the best way to manage multiples connectors in one command
For example, I choose to add a `--all` option to process all connectors. But it could be a good idea to allow regexp or wildcard pattern instead (or both)
For apply, I choose to not allow `-n` with multiple `-f` because it can be tricky to use otherwise. It might be too restrictive.
Just let me know if it's going the good way. I can adapt the implementation following recommendations.
And then apply the changes on all the other commands where it applies.


src/main/java/org/kcctl/command/ApplyCommand.java
```diff
@@ -33,19 +33,22 @@
...
 import static org.kcctl.util.Colors.ANSI_RESET;
 import static org.kcctl.util.Colors.ANSI_WHITE_BOLD;
 
 @Command(name = "apply", description = "Applies the given file or the stdin content for registering or updating a connector")
 public class ApplyCommand implements Callable<Integer> {
 
-    @Option(names = { "-f", "--file" }, description = "Name of the file to apply or '-' to read from stdin", required = true)
-    File file;
+    @Option(names = { "-f",
+            "--file" }, description = "Name of the file to apply or '-' to read from stdin. You can use multiple -f if the connectors names are in the files", required = true)
```
gunnarmorling, line 50:
```suggestion
            "--file" }, description = "Name of the file to apply or '-' to read from stdin. You can specify -f multiple times, if the connector names are in the files", required = true)
```

src/main/java/org/kcctl/command/ApplyCommand.java
```diff
@@ -68,17 +71,42 @@ public ApplyCommand(ConfigurationContext context) {
...
+        for (File file : files) {
+            returnCode = max(returnCode, applyOrValidateConnector(kafkaConnectApi, file));
+        }
+        return returnCode;
+    }
+
+    private void validate() {
+        if (files.length > 1 && name != null)
+            throw new CommandLine.ParameterException(spec.commandLine(),
+                    "It is not possible to use -n with multiple -f. Please provide the connector name in the connector configuration file.");
```
gunnarmorling, line 88:
```suggestion
                    "It is not possible to use -n when -f is given multiple times. Please provide the connector names in the connector configuration files.");
```

src/main/java/org/kcctl/command/ApplyCommand.java
```diff
@@ -68,17 +71,42 @@ public ApplyCommand(ConfigurationContext context) {
...
+
+    private String readConfig(File file) {
         String contents;
         if (file.getName().equals("-")) {
             contents = readFromStdin();
         }
         else if (!file.exists()) {
-            spec.commandLine().getOut().println("Given file does not exist: " + file.toPath().toAbsolutePath());
-            return 1;
+            throw new CommandLine.ExecutionException(spec.commandLine(), "Given file does not exist: " + file.toPath().toAbsolutePath());
```
gunnarmorling, line 109: What's the reason for raising an exception rather than returning a none 0 value?

src/main/java/org/kcctl/command/PauseConnectorCommand.java
```diff
@@ -26,11 +28,13 @@
 import picocli.CommandLine.Command;
 import picocli.CommandLine.Parameters;
 
-@Command(name = "connector", description = "Pauses the specified connector")
+@Command(name = "connector", aliases = "connectors", description = "Pauses the specified connector")
```
gunnarmorling, line 31:
```suggestion
@Command(name = "connector", aliases = "connectors", description = "Pauses the specified connector(s)")
```

gunnarmorling (commented):
Hey @jvenant, this looks great! Some minor comments/suggestions inline.

> For example, I choose to add a --all option to process all connectors. But it could be a good idea to allow regexp or wildcard pattern instead (or both)

Good question; in general, I prefer if there's one way to acomplish a given task, not multiple ones. So if you foresee the need for selecting names using regexps, I wouldn't add `--all`, because that'd be the same as `--name-pattern=.*`. Given regexps are more flexible and "all" is just a special case, I'd probably go for the more powerful variant and add `--name-pattern`.

> For apply, I choose to not allow -n with multiple -f because it can be tricky to use otherwise. It might be too restrictive.

I think that's very sensible. Let's do it that way.

> Just let me know if it's going the good way. I can adapt the implementation following recommendations.
> And then apply the changes on all the other commands where it applies.

+1 I think you're on the right track. Thanks a lot!

src/main/java/org/kcctl/command/ApplyCommand.java
jvenant: Good question, I asked to myself if it was a good idea.
The pragmatic answer is because the method return a String ^_^
But it can be manage of course
The underneath question is more global. It is about how to manage error when multiple connector are processed.

- Should we run everything and then report all errors at the end ?
- Should we stop on first error

But for the `readConfig` specific case, maybe it would be smarter to first read all the files and then only if evrything is ok, launch the apply process.

What do you thing ?

jvenant:
for the regexp stuff : I was thinking about a `-e` option that would then treat the names as regexp filter on all connectors names instead of `--name-pattern=.*`.
what do you think ?

jvenant:
I'm facing difficulties generating the completion. Picocli seems to (currently) require an empty constructor. Which is not compatible with the constructror injection pattern used on some command

gunnarmorling:
> Picocli seems to (currently) requires an empty constructor. Which is not compatible with the constructror injection pattern used on some commands

Yes, that's true. I thought I'd have added those default constructors to all the commands, but I may have missed some. Just add them in this case. They'll only be used for the completion generation, otherwise the inject-ed constructors are used.

gunnarmorling:
> for the regexp stuff : I was thinking about a -e option that would then treat the names as regexp filters on all connectors names (like grep) instead of --name-pattern=.*.

So would -e take the pattern then (which I think grep is doing)? Or would -e be a "switch" that impacts how the --name value is interpreted? Can you share the entire command invocation perhaps?

src/main/java/org/kcctl/command/ApplyCommand.java
gunnarmorling: > But for the readConfig specific case, maybe it would be smarter to first read all the files and then only launch the apply process if evrything is ok.

Yes, I think that makes sense. In general, I would

* Try to do as much validation/verification before starting mutating actions
* Abort after the first failure of a mutating action

jvenant:
> > for the regexp stuff : I was thinking about a -e option that would then treat the names as regexp filters on all connectors names (like grep) instead of --name-pattern=.*.
> 
> So would -e take the pattern then (which I think grep is doing)? Or would -e be a "switch" that impacts how the --name value is interpreted? Can you share the entire command invocation perhaps?

Good point : Not exactly like grep then. I see it more like a switch ;
`kcctl pause connector -e 'topic\.\d' 'othertopic\.\d'`
vs
`kcctl pause connector topic.1 topic.2 othertopic.1 othertopic.2 ...`


It could be confusing as the user may consider the `-e` as a tag for the name just after. But I feel like it's more compliant with the way the commands are currently designed.
tell me your preference : 
- `--name-pattern` before 'normal' names : then we have to decide how to manage mix of `--name-pattern` and 'normal' names argument
- `-e` like grep : might be tricky to manage commands like ` -e 'reg1' 'name1' -e 'reg2' 'name2'` 
- `-e` like a switch

jvenant:
> > Picocli seems to (currently) requires an empty constructor. Which is not compatible with the constructror injection pattern used on some commands
> 
> Yes, that's true. I thought I'd have added those default constructors to all the commands, but I may have missed some. Just add them in this case. They'll only be used for the completion generation, otherwise the inject-ed constructors are used.

I'll do

jvenant:
I made a commit taking in account your feedback. Currently it is a switch -e on pause, but I can change to something else.

I redesign the patch for multiple files on Apply. I think it is more robust like this, but it involve more changes in the code that I was expecting. So don't hesitate to tell me if you're not OK with some of my changes. I can adapt.

I'll apply the changes on others commands once we are OK with this ones

gunnarmorling:
> Currently it is a switch -e on pause, but I can change to something else.

Cool, I think that's a good approach. I wa leaning towards `--name-pattern` initially, since I (incorrectly) assumed that the connector name would currently be given via `--name`. But instead it's just given after the command (i.e. a parameter and not an option in PicoCli terms). So having an option like --regexp which controls how the parameter is interpreted makes sense 👍 .

My only question would be why "-e"? I.e. what does "e" it stand for, "expression"? How about using "-r"/"--reg-exp" for easier memorization? Btw. on "--reg-exp" vs. "--regexp", it probably should be the former, as we use the  dash in multi-word option names (e.g. see [SetContextCommand](https://github.com/kcctl/kcctl/blob/main/src/main/java/org/kcctl/command/SetContextCommand.java)).

gunnarmorling:
> I redesign the patch for multiple files on Apply. I think it is more robust like this, but it involve more changes in the code that I was expecting. So don't hesitate to tell me if you're not OK with some of my changes. I can adapt.

Thanks, it all LGTM and I think we've achieved agreement on the way forward. Apart from bringing regexp support to all commands taking a connector or task name, one other thing to do would be updating the list of commands in the README (it's the out of kcctl without any commands).

jvenant:
Patch already have a `-r` option

jvenant:
I currently setup `-r` and `--reg-exp` on all commands except Patch who is `-e` `--reg-exp`
probably not the best solution
Let me know how you want to deal with the `-r` short option conflict

gunnarmorling:
> Patch already have a -r option

Ah yes, I see. So looks like we should use `-e` / `--reg-exp` consistently for all commands then. Using different short names for the same option for different commands is too inconsistent and confusing.

gunnarmorling:
Interesting, I haven't seen that one before. Are you using Java 17 to build
kcctl=

jvenant:
Yes java 17. It works without my commit. 
I'll dig then

jvenant:
It's related to the ApplyConnector record I added in ApplyCommand.
Graalvm doesn't seem to like when it is part of a class field type
So I changed the `applyConnectors` type from `ApplyConnector[]` to `List` (changing the visibility of the record doesn't help. `List<ApplyConnector>` neither)
It's a little bit uggly as it force to cast. But it should do the job

So : I think we're good...
Tell me if you see anything

gunnarmorling:
Hey @jvenant, so this LGTM, merging. Thanks a lot for this very nice piece of work!

Issue #182, opened 2022-05-24
Want to sort the connector config info when executing the describe command


harveyyue:
kcctl describe connector  test_avro_source_v2

Name:       test_avro_source_v2
Type:       source
State:      RUNNING
Worker ID:  127.0.0.1:8083
Config:  
  connector.class:                           io.debezium.connector.mysql.MySqlConnector
  database.history.kafka.bootstrap.servers:  localhost:9092
  database.history.kafka.topic:              history_test_avro
  database.hostname:                         localhost
  database.include.list:                     workdb,demo
  database.password:                         ${sechub:15f99c3a735e6101ea14f63173914301dadce709a80272fcefe662cd40bcc904:LTyD3uSno/34YSN7XishKwbZUPr00/+KxO6emAwcCn8fynI=}
  database.port:                             3307
  database.server.name:                      test_avro
  database.user:                             mysqluser
  decimal.handling.mode:                     string
  include.schema.changes:                    true
  include.schema.comments:                   true
  key.converter:                             io.confluent.connect.avro.AvroConverter
  key.converter.schema.registry.url:         http://localhost:8081
  name:                                      test_avro_source_v2
  producer.override.batch.size:              1024000
  producer.override.buffer.memory:           40960000
  producer.override.linger.ms:               50
  producer.override.max.request.size:        4096000
  producer.override.retry.backoff.ms:        200
  snapshot.locking.mode:                     none
  snapshot.mode:                             schema_only
  tasks.max:                                 1
  value.converter:                           io.confluent.connect.avro.AvroConverter
  value.converter.schema.registry.url:       http://localhost:8081
Tasks:  
  0:  
    State:      RUNNING
    Worker ID:  127.0.0.1:8083
Topics:  
  test_avro.workdb.sys_dict_item
  test_avro

gunnarmorling:
Hey @harveyyue, nice to see you here! Can you elaborate a bit, what exactly would you like to have sorted, in which way? Note there's issue #6, which discusses grouping the output of the `kcctl describe` command; is that what you had in mind?

harveyyue:
Hi @gunnarmorling, haven't saw you for some time in debeizum community, in this case, I just want to make these properties  readable, we use ssl kafka with more relation ssl properies, it's hard to find some one from these mixed propeties.

gunnarmorling:
> Hi @gunnarmorling, haven't saw you for some time in debeizum community

Hehe, yeah, I'm focusing on another project these days.

[...]

PR #183, opened 2022-05-24
#182 Want to sort the connector config info when executing the describe command

harveyyue:
https://github.com/kcctl/kcctl/issues/182


gunnarmorling:
Applied, thx!

Issue #184, opened 2022-05-25
Introduce a new "expand" api to request all connectors status and config info


harveyyue:
The expand url example:
localhost:8083/connectors/expand=status&expand=info
localhost:8083/connectors/expand=status
localhost:8083/connectors/expand=info

The status expand response json:

````
{
    "test_avro_cache5_source_v2": {
        "status": {
            "name": "test_avro_cache5_source_v2",
            "connector": {
                "state": "RUNNING",
                "worker_id": "127.0.0.1:8083"
            },
            "tasks": [
                {
                    "id": 0,
                    "state": "RUNNING",
                    "worker_id": "127.0.0.1:8083"
                }
            ],
            "type": "source"
        }
    },
    "test_avro_strategy_source": {
        "status": {
            "name": "test_avro_strategy_source",
            "connector": {
                "state": "RUNNING",
                "worker_id": "127.0.0.1:8083"
            },
            "tasks": [
                {
                    "id": 0,
                    "state": "RUNNING",
                    "worker_id": "127.0.0.1:8083"
                }
            ],
            "type": "source"
        }
    },
    "test_avro_source_v2": {
        "status": {
            "name": "test_avro_source_v2",
            "connector": {
                "state": "RUNNING",
                "worker_id": "127.0.0.1:8083"
            },
            "tasks": [
                {
                    "id": 0,
                    "state": "RUNNING",
                    "worker_id": "127.0.0.1:8083"
                }
            ],
            "type": "source"
        }
    }
}
````

gunnarmorling:
Closing as done, forgot about that before. Thx, @harveyyue!

PR #185, opened 2022-05-25
#184 Introduce a new "expand" api to request all connectors status and config info

harveyyue:
https://github.com/kcctl/kcctl/issues/184

### DIFF
--- a/src/main/java/org/kcctl/service/TaskInfo.java
+++ b/src/main/java/org/kcctl/service/TaskInfo.java
@@ -15,8 +15,5 @@
  */
 package org.kcctl.service;
 
-public class TaskInfo {
-
-    public String connector;
-    public Integer task;
+public record TaskInfo(String connector, Integer task) {
 }
--- a/src/main/java/org/kcctl/service/TaskInfo.java
+++ b/src/main/java/org/kcctl/service/TaskInfo.java
@@ -15,13 +15,8 @@
  */
 package org.kcctl.service;
 
-import java.util.List;
-import java.util.Map;
+public class TaskInfo {
 
-public class ConnectorInfo {
-
-    public String name;
-    public Map<String, String> config;
-    public List<TaskInfo> tasks;
-    public String type;
+    public String connector;
+    public Integer task;
 }
--- a/.github/workflows/early-access.yml
+++ b/.github/workflows/early-access.yml
@@ -41,7 +41,7 @@ jobs:
 
       - uses: graalvm/setup-graalvm@v1
         with:
-          version: '22.1.0.1'
+          version: '22.1.0'
           java-version: '17'
           components: 'native-image'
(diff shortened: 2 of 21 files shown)


gunnarmorling:
Hey @harveyyue, so how exactly will the output of `kcctl get connectors` differ with this change? Also, shouldn't the "expand" attribute be a command parameter somehow?

harveyyue:
@gunnarmorling 

> so how exactly will the output of kcctl get connectors differ with this change? 
Expose the worker_id including ip and port to TASKS, since we deployed the connect service with cluster mode, this is help us to known the connector location.

The command `kcctl get connectors` ouput example:
`
harvey_test_topic_strategy_jdbc_sink          sink        RUNNING   0: some_ip:28083 FAILED
harvey_test_sechub_v2_source                     source   RUNNING   0: some_ip:28083 RUNNING
`

> Also, shouldn't the "expand" attribute be a command parameter somehow?
Yes, only add a "expand" api to get all connectors status/configs instead of getting these info one by one with related api.

gunnarmorling:
Aaah, so I think I get it now. This doesn't change at all how the command
behaves to the user, it "only" implements it more efficiently by doing just
a single call to Connect, right?

That makes sense; we'd need to check the backend version though in order to
make sure we talk to a version new enough so it actually supports that new
call. Otherwise, we'd have to fall back to the existing implementation.

gunnarmorling:
Here's another case btw. where we check the version of Kafka Connect and take action based on that: https://github.com/kcctl/kcctl/blob/main/src/main/java/org/kcctl/command/DescribeConnectorCommand.java#L74-L83.

harveyyue:
Connect introduce this "expand" feature since 2.3.0-rc1. Yes, we need to add this version switch to invoke the corrected api.
[ConnectorsResource](https://github.com/apache/kafka/blob/2.3.0-rc1/connect/runtime/src/main/java/org/apache/kafka/connect/runtime/rest/resources/ConnectorsResource.java)

src/main/java/org/kcctl/service/TaskInfo.java
```diff
@@ -0,0 +1,22 @@
...
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package org.kcctl.service;
+
+public class TaskInfo {
```
gunnarmorling, line 18: Could be a record type, given you started using them elsewhere? Same for a few other types.

gunnarmorling:
LGTM, thanks a lot, @harveyyue! Anything else you wanna push to this PR before I merge it?

gunnarmorling:
Applied; thx, @harveyyue!

PR #195, opened 2022-05-26
🚀 feat(apply): add parametric json configuration support

muhammedikinci:
As a developer: I don't want to put my connector configs in the company repository with database passwords or sensitive data.

Description: Pass parameters to config for sensitive data when creating and updating the connector.

I add a new argument `--config-parameters` to `apply` command.

Let's say we have a connector config like the below. If I put this configuration in the repository, this does not allow for security. I can set sensitive parts to blank for the repository but if I put this configuration to pipeline or automation for generic configuration that will not work I must create many same configuration file just for one field and this is not cool.
```json
{
  "name": "my-connector",
  "config": {
    "connector.class": "io.debezium.connector.mysql.MySqlConnector",
    "database.user": "debezium-user",
    "database.password": "password",
    ...
  }
}
```

So, I can put parameters to these fields and I can set them from the command line
```json
{
  "name": "my-connector",
  "config": {
    "connector.class": "io.debezium.connector.mysql.MySqlConnector",
    "database.user": "$#1",
    "database.password": "$#2",
    ...
  }
}
```

And I can run this command;
```sh
kcctl apply -f test-connector.json -ca "debezium-user password"
```

Or I can use for duplicate fields
```json
{
  ...
  "key.converter": "io.confluent.connect.avro.AvroConverter",
  "key.converter.schema.registry.url": "$#1",
  "value.converter": "io.confluent.connect.avro.AvroConverter",
  "value.converter.schema.registry.url": "$#1"
  ...
}
```

### DIFF
--- a/src/main/java/org/kcctl/command/ApplyCommand.java
+++ b/src/main/java/org/kcctl/command/ApplyCommand.java
@@ -70,6 +70,9 @@ public class ApplyCommand implements Callable<Integer> {
     @Spec
     CommandSpec spec;
 
+    @Option(names = { "-ca", "--config-parameters" }, description = "Pass parameters to config for sensitive data")
+    String configParameters;
+
     private static final ObjectMapper mapper = new ObjectMapper();
     private final static String CONFIG_EXCEPTION = "org.apache.kafka.common.config.ConfigException: ";
     private final ConfigurationContext context;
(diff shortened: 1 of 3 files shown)


gunnarmorling:
Hey @muhammedikinci, sorry for the late feedback, only seeing this now. It's an interesting proposal, I'm curious though what advantages you see over the secret provider mechanism built into Kafka Connect itself? In particular, when using that, you also won't expose any secrets when displaying connector configuration.

muhammedikinci:
Thank you for your interest. You are right, actually, I didn't see the secret provider. It resolves all these problems.

Issue #204, opened 2022-06-25
The describe command meet a not existed connector and get the NPE instead of the available connectors


harveyyue:
Connector x not found. The following connector(s) are available:
java.lang.NullPointerException
	at org.kcctl.command.GetConnectorsCommand.run(GetConnectorsCommand.java:77)
	at org.kcctl.command.DescribeConnectorCommand.describe(DescribeConnectorCommand.java:183)
	at org.kcctl.command.DescribeConnectorCommand.call(DescribeConnectorCommand.java:94)
	at org.kcctl.command.DescribeConnectorCommand.call(DescribeConnectorCommand.java:47)
	at picocli.CommandLine.executeUserObject(CommandLine.java:1953)
	at picocli.CommandLine.access$1300(CommandLine.java:145)
	at picocli.CommandLine$RunLast.executeUserObjectOfLastSubcommandWithSameParent(CommandLine.java:2358)
	at picocli.CommandLine$RunLast.handle(CommandLine.java:2352)
	at picocli.CommandLine$RunLast.handle(CommandLine.java:2314)
	at picocli.CommandLine$AbstractParseResultHandler.execute(CommandLine.java:2179)
	at picocli.CommandLine$RunLast.execute(CommandLine.java:2316)
	at io.quarkus.picocli.runtime.PicocliRunner$EventExecutionStrategy.execute(PicocliRunner.java:26)
	at picocli.CommandLine.execute(CommandLine.java:2078)
	at io.quarkus.picocli.runtime.PicocliRunner.run(PicocliRunner.java:40)
	at io.quarkus.runtime.ApplicationLifecycleManager.run(ApplicationLifecycleManager.java:124)
	at io.quarkus.runtime.Quarkus.run(Quarkus.java:67)
	at io.quarkus.runtime.Quarkus.run(Quarkus.java:41)
	at io.quarkus.runner.GeneratedMain.main(Unknown Source)

gunnarmorling:
Fixed by #205. Thx, @harveyyue!

PR #205, opened 2022-06-25
#204 The describe command meet a not existed connector and get the NPE instead of the available connectors

harveyyue:


gunnarmorling:
Good spot, thx!

Issue #207, opened 2022-06-29
Add support for KIP-769


mimaison:
[KIP-769](https://cwiki.apache.org/confluence/display/KAFKA/KIP-769%3A+Connect+APIs+to+list+all+connector+plugins+and+retrieve+their+configuration+definitions) allows listing all connector plugins and also retrieving their configuration definitions. kcctl should expose these features.

I've started working on it, hoping to have a PR soon

gunnarmorling:
Woot, woot. Looking forward to the PR. How were you thinking to expose this, via `kcctl describe plug-in SomePlugin`?

gunnarmorling:
Sounds great!

> `kcctl get plugins --all` to list all plugins

That would work; alternatively, how about `kcctl get plugins --type={connector|converter|transformation|predicate|all}? I.e. providing some more fine-grained control over the returned plug-in types?

gunnarmorling:
Hey @mimaison, I think this one is actually done, or is there any remaining task here?

mimaison:
Yes this is complete, closing.

PR #230, opened 2022-07-26
🔍 Add support for KIP-769 features

mimaison:
- `describe plugin <plugin>` to retrieve a plugin's configuration definitions
- `get plugins --types=<types>` to list plugins by types


gunnarmorling:
Fixes #207.

src/main/java/org/kcctl/command/GetPluginsCommand.java
```diff
@@ -33,13 +37,19 @@
...
+
+    private final Version requiredVersionForAllPlugins = new Version(3, 2);
+    private final Set<String> defaultPlugins = Set.of("source", "sink");
 
     private final ConfigurationContext context;
 
     @CommandLine.Spec
     CommandLine.Model.CommandSpec spec;
 
+    @CommandLine.Option(names = { "-t", "--types" }, description = "Types of plugins", split = ",")
```
gunnarmorling, line 50: Can we make this an [enum option](https://picocli.info/#_enum_types)? That way, we get validation of the passed value(s), and the allowed types will be available for tab completion, too.

src/main/java/org/kcctl/command/GetPluginsCommand.java
```diff
@@ -33,13 +37,19 @@
 import picocli.CommandLine.Command;
 
 @Command(name = "plugins", description = "Displays information about available connector plug-ins")
-public class GetPluginsCommand implements Runnable {
+public class GetPluginsCommand implements Callable<Integer> {
+
+    private final Version requiredVersionForAllPlugins = new Version(3, 2);
+    private final Set<String> defaultPlugins = Set.of("source", "sink");
```
gunnarmorling, line 43: How about getting all plug-in types by default? I think the backwards compatibility concern is less important for kcctl given its current pre 1.0 state.

src/main/java/org/kcctl/command/GetPluginsCommand.java
```diff
@@ -51,19 +61,35 @@ public GetPluginsCommand() {
...
 
-        List<ConnectorPlugin> connectorPlugins = kafkaConnectApi.getConnectorPlugins();
-        connectorPlugins.sort((c1, c2) -> -c1.type().compareTo(c2.type()));
+        Version currentVersion = new Version(kafkaConnectApi.getWorkerInfo().version());
+
+        if (pluginTypes != null) {
+            if (!currentVersion.greaterOrEquals(requiredVersionForAllPlugins)) {
+                spec.commandLine().getErr().println("--types requires at least Kafka Connect 3.2. Current version: " + currentVersion);
+                return 1;
+            }
```
gunnarmorling, line 75: We could apply client-side filtering of the result set for KC < 3.2. We have similar logic elsewhere already I think, kinda emulating features when absent in older versions and doable with a reasonable effort.

src/main/java/org/kcctl/command/DescribePluginCommand.java
```diff
@@ -0,0 +1,85 @@
...
+import picocli.CommandLine;
+import picocli.CommandLine.Command;
+
+@Command(name = "plugin", description = "Displays information about given plugin")
+public class DescribePluginCommand implements Callable<Integer> {
+
+    @CommandLine.Spec
+    CommandLine.Model.CommandSpec spec;
+
+    @CommandLine.Parameters(paramLabel = "PLUGIN NAME", description = "Name of the plugin")
```
gunnarmorling, line 43: Let's implement a tab completion of the name; it will be particularly handy as the plug-in names are so long typically. Unlike "static" completions for option names, this requires to invoke kcctl from the completion script, retrieving the available plug-ins. You can find an example [here](https://github.com/kcctl/kcctl/blob/main/src/main/java/org/kcctl/command/DescribeConnectorCommand.java#L53) (essentially, that custom completion retrieves the names by calling another command which is hidden from the help text and completion itself).

Then regenerate the actual completion script itself, as described [here](https://github.com/kcctl/kcctl#updating-the-completion-script) (note the special backtick handling for that additional new _kcctl_ invocation). This will also automatically add the completion handling for the new types option and its allowed values.

gunnarmorling (commented):
Very nice; thanks a lot, @mimaison! Added a few comments inline, mostly centered around tab completion handling.

src/main/java/org/kcctl/command/GetPluginsCommand.java
mimaison: Good idea, updated!

src/main/java/org/kcctl/command/GetPluginsCommand.java
mimaison: The issue is that pre 3.2 you cannot list all plugins. I've updated the logic to return an error in case you try to list plugins other than sink and source without 3.2.

For example:
```
> kcctl get plugins --types=predicate
Listing plugins other than source and sink requires at least Kafka Connect 3.2. Current version: 3.1
> kcctl get plugins --types=sink

 TYPE   CLASS                                                   VERSION
 sink   org.apache.kafka.connect.file.FileStreamSinkConnector   3.1.0
```

src/main/java/org/kcctl/command/GetPluginsCommand.java
mimaison: I've updated the option to use an enum. The completion file has been updated and looks fine but I can't get the completion to work when I type `kcctl get plugins --types=<TAB>` so not sure if it's correct.

src/main/java/org/kcctl/command/DescribePluginCommand.java
mimaison: Do we need to do this in addition of using an enum? I guess this way it would only show types of plugins installed instead of the full list of types but I'm not sure this adds a lot of value.

mimaison:
Thanks @gunnarmorling for the review. I've pushed an update.

src/main/java/org/kcctl/command/GetPluginsCommand.java
gunnarmorling: Interesting, so that does work for me:

```
kcctl get plugins --types
converter         header_converter  predicate         sink              source            transformation
```

Where it falls short is when specifying multiple values, as completion won't work after a comma: `kcctl get plugins --types source,<TAB>`, but I think that's ok.

src/main/java/org/kcctl/command/DescribePluginCommand.java
gunnarmorling: I was thinking of a completion of the plug-in _name_, not type, for the describe command; e.g. like so:

```
kcctl describe plugin io.debezium<TAB>
io.debezium.postgres.PostgresConnector...
```

That way, you don't have to type out the FQNs for the describe command. I have pushed a commit for adding that. Perhaps you can give it a spin and let me know whether that completion works for you as it should?

gunnarmorling:
Hey @mimaison, this LGTM. I've pushed one more commit for adding the completion on plug-in names.

mimaison:
Thanks @gunnarmorling for the feedback!

I've adjusted the output so it prints plugin configurations as you suggested, I agree it's more readable for plugins with long values. I've reused `Tuple` and noticed it writes to `System.out` while some commands write to `spec.commandLine().getOut()`. I tried quickly to clean this up (and use `spec.commandLine().getOut()` everywhere) but ran into an issue with `PatchConnectorCommand` which invokes another command and that loses `spec`.

Regarding the issue you saw with plugins with `$`, it's caused by the shell. So you need to either quote the plugin name or escape the dollar sign:
```
kcctl describe plugin 'org.apache.kafka.connect.transforms.ExtractField$Key'
kcctl describe plugin org.apache.kafka.connect.transforms.ExtractField\$Key
```
I'm not sure what can be done to change that. One option that feels a bit too drastic would be to display slightly adjusted names in kcctl output for these for example `ExtractField-Key` but it's pretty risky, ie what happens if a plugin actually has a dash in its name.

gunnarmorling:
Very nice, rebased and applied. Thanks a lot, @mimaison! The output looks much better now. Regarding the "$", thanks for investigating that. I think it's fine as-is for now, as you say, it works when putting the class name in quotes.

Issue #253, opened 2022-09-29
Set max width value for tasks column of GetConnectorsCommand


harveyyue:
Just want to adjust the column width 100 instead of default value 80, to avoid printing too crowded when some connector have multiple tasks.

PR #254, opened 2022-09-29
#253 Set max width value for tasks column of GetConnectorsCommand

harveyyue:


gunnarmorling:
Sorry for the delay, somehow missed this one. LGTM, applied. Thanks, @harveyyue!

Issue #261, opened 2022-10-13
Provide option to filter output of kcctl describe plugin


gunnarmorling:
Via `kcctl describe plugin` one can list all the options of a connector (or any other plug-in like SMTs). For connectors with a large config surface, such as Debezium's, the output can be a bit overwhelming though. It could be useful to be able to filter the output, e.g. when knowing the name of one option, or some word from its description. I thought of supporting `kcctl describe plugin --grep <some search string>`, but am open for other proposals too.

gunnarmorling:
@C0urante, @mimaison, interesting for you folks perhaps?

C0urante:
Yeah, this seems useful.

There are a few decisions we'd need to make:
1. Total search space: the issue description mentions searching by name and description. Would we want to also allow searching by other characteristics such as importance, group, or display name?
2. Configurable search space: would it make sense to hardcode search to use every possible space (name, description, etc.), or would we want to allow people to narrow the search space exclusively to, e.g., name or description?
3. Search syntax: the `--grep` flag implies regex search; other options also include literal substring search and shell glob search.
4. Number of terms: should we support only a single search term, or multiple? If multiple, should they be joined via logical "and" or logical "or"?

And here are my initial thoughts on each:
1. Name and description seem like a good start. Display name and group may be useful for programmatic inspection, but IIUC they're mostly there for UIs to use, and not really intended for humans to work with directly. Filtering by importance could be useful too, but I'm not sure we need it for the first iteration of this feature. If possible we should design the API in a way that allows us to evolve on this front in the future.
2. It could be pretty useful to limit the search space to just property names. Someone might want to know all database-related properties for the Debezium MySQL connector; if they search for the term `database`, the results could be pretty noisy if we include everything whose name or description matches that term. Not sure if the alternative case (searching exclusively by description) would ever come in handy, so we don't necessarily have to add support for that if we can't do it without sacrificing API cleanliness. To help make the matching behavior clear to users, we may want to highlight the portion of each property that matched the search.
[...]

PR #263, opened 2022-10-14
#261 Add search filter API for plugin descriptions

C0urante:
Addresses https://github.com/kcctl/kcctl/issues/261

Adds support for the mutually-exclusive flags `--search`, `--search-name`, and `--search-description` to the `kcctl describe plugin` command, including automatic underlining of matches.

I opted to underline instead of bold, highlight, change color, etc. since with my limited prototyping it seemed the most readable option and least likely to interact poorly with terminals whose color schemes differ from mine (I use white font on a black background). Would be interested if anyone has any alternatives to propose.

### DIFF
--- a/src/main/java/org/kcctl/util/MutableConfigKeyInfo.java
+++ b/src/main/java/org/kcctl/util/MutableConfigKeyInfo.java
@@ -1,0 +1,199 @@
+/*
+ *  Copyright 2021 The original authors
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package org.kcctl.util;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Objects;
+
+import static org.kcctl.service.ConfigInfos.ConfigKeyInfo;
+
+public class MutableConfigKeyInfo {
+
+    private String name;
+    private String type;
+    private boolean required;
+    private String defaultValue;
+    private String importance;
+    private String documentation;
+    private String group;
+    private int orderInGroup;
+    private String width;
+    private String displayName;
+    private List<String> dependents;
+
+    public MutableConfigKeyInfo(
+                                String name,
+                                String type,
+                                boolean required,
+                                String defaultValue,
+                                String importance,
+                                String documentation,
+                                String group,
+                                int orderInGroup,
+                                String width,
+                                String displayName,
+                                List<String> dependents) {
+        this.name = name;
+        this.type = type;
+        this.required = required;
+        this.defaultValue = defaultValue;
+        this.importance = importance;
+        this.documentation = documentation;
+        this.group = group;
+        this.orderInGroup = orderInGroup;
+        this.width = width;
+        this.displayName = displayName;
+        this.dependents = dependents;
+    }
+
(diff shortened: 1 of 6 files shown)


src/main/java/org/kcctl/util/Search.java
```diff
@@ -0,0 +1,116 @@
...
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import static org.kcctl.service.ConfigInfos.ConfigKeyInfo;
+
+public class Search {
+
+    // Generic records... I'm going to hell for this
+    public record SearchTerm<E>(Function<E, String> extractTerm, BiConsumer<E,String>setTerm)
+    {
```
gunnarmorling, line 32: Hum, wondering why the formatter doesn't flag this. The opening brace should at the end of the line above, the matches method should be intended.

gunnarmorling (commented):
I think the proposed implementation looks great; some minor comments inline; and a few more questions here:

* Could you update the completion file, please (see description in the README)?
* How can one achieve a case-insensitive search? Probably worth adding that to the option descriptions, too.
* In regards to the result highlighting, I feel the underline style is rather... subtle. I get the concern about coloring, how about using `ANSI_WHITE_BOLD` though (in addition to underlining), as already is in use for the "Name:", "Type:", etc. literals? I suppose your shell color theme would automatically invert that? Alternatively, how about using a [background color](TRUNCATE)?

src/main/java/org/kcctl/util/Search.java
C0urante: This is actually the result of the autoformatter; I fixed the styling and then reran `./mvnw process-sources` and it reverted to what's in the PR.

I tried to do some digging into what's causing this but haven't been able to find the part of the formatter (or its dependencies) that's responsible.

src/main/java/org/kcctl/util/MutableConfigKeyInfo.java
```diff
@@ -0,0 +1,199 @@
...
+ */
+package org.kcctl.util;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Objects;
+
+import static org.kcctl.service.ConfigInfos.ConfigKeyInfo;
+
+public class MutableConfigKeyInfo {
```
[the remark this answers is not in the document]
C0urante: 👍 Ack, added.

src/main/java/org/kcctl/util/Search.java
gunnarmorling: I see; kinda weird. Did you try updating to the latest version (if there is any newer)? Probably not worth spending too much time on it, at least it's formatted consistently like that, so no friction during any future PRs.

gunnarmorling:
Hey @C0urante, so IMO it's much better with the background highlighting. I think we can merge this now. Btw. there already is an issue with color schemes which don't use a dark background; at least when I change to "Light Background" preset in iterm2 on macOS, the bold white color isn't inverted (the regular light white text is), i.e. the literals like "Name:" etc. in the output of this command are not readable. Not sure what could be done about this, one could argue it's an issue in the color preset of iterm2 which should take care of that.

gunnarmorling:
Rebased, squashed and applied. Thanks a million, @C0urante! A really great contribution 🙏 !

C0urante:
Thanks @gunnarmorling! Appreciate the rapid turnaround. Any timeline on the 1.0.0 release?

gunnarmorling:
> Thanks @gunnarmorling! Appreciate the rapid turnaround.

Thank _you_ really for putting in the effort! Just in case you're looking for more, I logged a few more issues over the weekend ;)

> No rush, but curious--any timeline on the 1.0.0 release?

Good question, haven't thought a lot about this. I didn't feel the need, given the availability of the EA releases after each push to `main`. But we could cut that release soon-ish, if you think it would instill more trust by potential users? I don't think anything important is missing feature-wise at this point. When doing so, we should do take a look at the existing commands and options and consistency though, for instance the `delete` inconsistency should be fixed first. After 1.0.0, we'd be a bit more limited in terms of changing and breaking things.

gunnarmorling:
Btw. I've just sent you an invite to join as a committer. Just if you're interested, no pressure whatsoever. In terms of what it means: you could review and merge PRs from others, e.g. me. We avoid pushing our own commits to _main_, apart from typo fixes and the like.

Issue #250, opened 2022-09-20
kcctl apply -f should allow multiple files


gunnarmorling:
Either -f should be repeatable, or it should allow for passing multiple files at once. The latter should work with expansion in shells, e.g. `kcctl apply -f *-connector.json`. // CC @mimaison; this just occurred to me when using kcctl during a demo the other day.

gunnarmorling:
Hey @C0urante, it was really cool to meet you at Current last week! As you indicated potential interest in contributing to kcctl and were asking for issues to pick up, this one could be a candidate.

C0urante:
Hey @gunnarmorling! Great to meet you as well :)

It seems like `-f` is already repeatable as of https://github.com/kcctl/kcctl/pull/167, so I'm guessing the point of this issue is to explore whether that should be the only way to allow multiple filenames.

The shell globbing example does a good job of demonstrating why having to prepend each filename with `-f ` is inconvenient.

One question that comes to mind is why we're using a flag for a required argument at all? When I gave `kcctl` a test drive I was surprised that I had to specify `-f` instead of just doing something like `kcctl apply my-connector.json`.

gunnarmorling:
> It seems like -f is already repeatable as of https://github.com/kcctl/kcctl/pull/167

Ha, completely forgot about that. Just verified it, this already works indeed.

>  why we're using a flag for a required argument at all?

This goes back to the project's tag line "like kubectl, but for Connect". In kubectl, you can do either -f (which could be a file, or a directory; we probably should support the latter too), or -k (for a Kustomize directory). I guess having support for Kustomize, or something in that spirit, could make sense eventually for kcctl too? In that light, keeping -f seems reasonable to me.

So on this issue itself, I see the following options:

1. We close it, as -f is repeatable already
2. We change it so that -f is not repeatable any longer, but -f allows to specify multiple files; this would implicitly work with filename globbing in the shell

My preference would be on option 2.

C0urante:
Hi @gunnarmorling--sorry for the delay! Option 2 sounds fine to me. I'll push something shortly.

gunnarmorling:
@C0urante Awesome, thanks a lot!

PR #281, opened 2022-11-28
#250 Allow multiple file names with 'apply -f'

C0urante:
Addresses #250 

Allows multiple files to be specified following a single `-f` flag, in order to work better with shell globbing.

I looked into the possibility of disallowing multiple `-f` flags from being specified, but based on https://github.com/remkop/picocli/issues/190 (which doesn't touch on that exact feature, but something related) and the picocli Javadocs, it appears that that's not supported.

### DIFF
--- a/src/main/java/org/kcctl/command/ApplyCommand.java
+++ b/src/main/java/org/kcctl/command/ApplyCommand.java
@@ -60,7 +60,7 @@ public class ApplyCommand implements Callable<Integer> {
     // Support "-f file1.json file2.json" in addition to "-f file1.json -f file2.json" in order to work smoothly with shell
     // globbing (e.g., "-f file*.json")
     @Option(names = { "-f",
-            "--file" }, description = "Names of the file to apply or '-' to read from stdin. You can specify multiple filenames, if the connector names are in the files", required = true, converter = ApplyConnectorConverter.class, arity = "1..*")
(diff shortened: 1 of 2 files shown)


src/main/java/org/kcctl/command/ApplyCommand.java
```diff
@@ -57,8 +57,10 @@ boolean isNamed() {
 
     // Hack : workaround. Should be `List<ApplyConnector>` instead of List.
     // But Graalvm seems to have difficulty building a native binary with ApplyConnector type record in class fields
+    // Support "-f file1.json file2.json" in addition to "-f file1.json -f file2.json" in order to work smoothly with shell
+    // globbing (e.g., "-f file*.json")
     @Option(names = { "-f",
-            "--file" }, description = "Name of the file to apply or '-' to read from stdin. You can specify -f multiple times, if the connector names are in the files", required = true, converter = ApplyConnectorConverter.class)
+            "--file" }, description = "Names of the file to apply or '-' to read from stdin. You can specify multiple filenames, if the connector names are in the files", required = true, converter = ApplyConnectorConverter.class, arity = "1..*")
```
gunnarmorling, line 63:
```suggestion
            "--file" }, description = "One or more name(s) of file(s) to apply, or '-' to read from stdin. You can specify multiple file names only if the connector names are in the files", required = true, converter = ApplyConnectorConverter.class, arity = "1..*")
```

gunnarmorling (approved):
LGTM. Did just one small wording change. Merging. Thanks, @C0urante!

C0urante:
Thanks Gunnar! Good catch.

Issue #265, opened 2022-10-16
Make delete command more regular


gunnarmorling:
Right now, the command to delete a connector is `kcctl delete <connector name>`. This is rather irregular, as for all the other command, the resource to act on is to be given as a sub-resource, e.g. `kcctl pause connector <connector name>`. Arguably, this is a bit more to type for those cases where there only is one kind of matching resource, but I still believe this is the better approach for the sake of regularity (e.g. `kcctl get ...` supports multiple resource types). I therefor suggest to apply the same reasoning to delete and change it into `kcctl delete connector <connector name>`.

gunnarmorling:
@C0urante, curious about your take.

C0urante:
Totally agree--it threw me the first time I tried to delete a connector that I didn't have to specify the resource type 👍

gunnarmorling:
Done.

PR #282, opened 2022-11-28
#265 replace 'delete' with 'delete connector'

C0urante:
Addresses #265

The `delete` command is replaced with `delete connector`. A useful error message is added to let users know about the replacement.

### DIFF
--- a/src/main/java/org/kcctl/command/DeleteCommand.java
+++ b/src/main/java/org/kcctl/command/DeleteCommand.java
@@ -1,0 +1,57 @@
+/*
+ *  Copyright 2021 The original authors
+ *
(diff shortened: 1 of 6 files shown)


gunnarmorling (approved):
LGTM, merging. This one was really bothering me ;) I think it might be about time to spin another release, so people can get this change via brew and SDKMan.

gunnarmorling:
Thx, @C0urante!

C0urante:
🎉 Nice!

Probably worth figuring out https://github.com/kcctl/homebrew-tap/pull/3 and any  SDKMan-equivalent issues before putting out a new release, but otherwise, yeah, we should publish this soon so that people have some time to get used to the new command syntax!

PR #288, opened 2022-12-20
adding a new option in describe to specify an output format, for now …

sebastienblanc:
adding a new option in the describe  command to specify an output format, for now just json
 `kcctl describe connector kafka-to-http --output-format json`

Fixed also a test that was using my actual kcctl context :)

### DIFF
--- a/src/main/java/org/kcctl/command/DescribeConnectorCommand.java
+++ b/src/main/java/org/kcctl/command/DescribeConnectorCommand.java
@@ -61,8 +61,8 @@ public class DescribeConnectorCommand implements Callable<Integer> {
     @Option(names = { "--tasks-config" }, description = "Displays tasks configuration")
     boolean includeTasksConfig;
 
-    @Option(names = { "--output-format" }, description = "Specifies the output format, i.e 'json'")
-    String outputFormat;
+    @Option(names = { "-o", "--output-format" }, description = "Specifies the output format, i.e 'json'")
+    OutputFormat outputFormat;
 
     private final ConfigurationContext context;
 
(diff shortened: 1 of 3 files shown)


src/main/java/org/kcctl/command/DescribeConnectorCommand.java
```diff
@@ -59,16 +61,24 @@ public class DescribeConnectorCommand implements Callable<Integer> {
     @Option(names = { "--tasks-config" }, description = "Displays tasks configuration")
     boolean includeTasksConfig;
 
+    @Option(names = { "--output-format" }, description = "Specifies the output format, i.e 'json'")
+    String outputFormat;
```
gunnarmorling, line 65: Can we make that an enum? picocli has bespoke support for enum options, e.g. in tab completion. Allowed values would be "text" and "json" for now.

gunnarmorling (commented):
Thank you so much, @sebastienblanc! It's a great start. Besides the inline comment, could you also:

* Add a test (see others for examples)
* Update the completion script, as described in the README

?

Thanks again!

sebastienblanc:
@gunnarmorling I've added a test and refactored to enum, regarding the completion script, I followed the instructions but I end up with a script with a LOT of changes, I'm not really feeling confident to commit it, can you advise me on it ?

gunnarmorling:
Superseded by https://github.com/kcctl/kcctl/pull/298, which I've just merged. Thanks a lot for this contribution, @sebastienblanc! I've pushed one more commit for updating the completion script; indeed this was a bit more involved than usually, I suppose it was caused by the Picocli version update. I've tested the completion script and it worked as expected, including your new option.

Issue #295, opened 2022-12-23
NullPointerException when trying to register a connector of an unknown type


gunnarmorling:
Ran into this when registering a connector whose type isn't available in that connect instance. The `spec` object is null in the programmatically created `GetPluginsCommand` object:

```
kcctl apply -f register-etcd.json
Specified class isn't a valid connector type. The following connector type(s) are available:
java.lang.NullPointerException: Cannot invoke "picocli.CommandLine$Model$CommandSpec.commandLine()" because "this.spec" is null
	at org.kcctl.command.GetPluginsCommand.call(GetPluginsCommand.java:84)
	at org.kcctl.command.ApplyCommand.createOrUpdateConnector(ApplyCommand.java:202)
	at org.kcctl.command.ApplyCommand.applyOrValidateConnector(ApplyCommand.java:161)
	at org.kcctl.command.ApplyCommand.call(ApplyCommand.java:142)
	at org.kcctl.command.ApplyCommand.call(ApplyCommand.java:43)
	at picocli.CommandLine.executeUserObject(CommandLine.java:2041)
	at picocli.CommandLine.access$1500(CommandLine.java:148)
	at picocli.CommandLine$RunLast.executeUserObjectOfLastSubcommandWithSameParent(CommandLine.java:2461)
	at picocli.CommandLine$RunLast.handle(CommandLine.java:2453)
	at picocli.CommandLine$RunLast.handle(CommandLine.java:2415)
	at picocli.CommandLine$AbstractParseResultHandler.execute(CommandLine.java:2273)
	at picocli.CommandLine$RunLast.execute(CommandLine.java:2417)
	at io.quarkus.picocli.runtime.PicocliRunner$EventExecutionStrategy.execute(PicocliRunner.java:26)
	at picocli.CommandLine.execute(CommandLine.java:2170)
	at io.quarkus.picocli.runtime.PicocliRunner.run(PicocliRunner.java:40)
	at io.quarkus.runtime.ApplicationLifecycleManager.run(ApplicationLifecycleManager.java:130)
	at io.quarkus.runtime.Quarkus.run(Quarkus.java:71)
	at io.quarkus.runtime.Quarkus.run(Quarkus.java:44)
	at io.quarkus.runner.GeneratedMain.main(Unknown Source)
	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:104)
[...]

PR #299, opened 2022-12-27
#295 Fix NPE for listing plugins when connector class not found

C0urante:
Addresses #295

### DIFF
--- a/src/test/java/org/kcctl/command/ApplyCommandTest.java
+++ b/src/test/java/org/kcctl/command/ApplyCommandTest.java
@@ -140,4 +140,14 @@ class ApplyCommandTest extends IntegrationTest {
         assertThat(context.output().toString()).doesNotContain("Created connector");
     }
 
+    @Test
+    public void should_gracefully_handle_missing_connector_class() {
+        var path = Paths.get("src", "test", "resources", "nonexistent.json");
+
+        int exitCode = context.commandLine().execute("-f", path.toAbsolutePath().toString());
+
+        assertThat(exitCode).isEqualTo(CommandLine.ExitCode.SOFTWARE);
+        assertThat(context.output().toString()).contains("Specified class isn't a valid connector type");
+    }
+
 }
--- a/src/main/java/org/kcctl/command/ApplyCommand.java
+++ b/src/main/java/org/kcctl/command/ApplyCommand.java
@@ -198,8 +198,11 @@ public class ApplyCommand implements Callable<Integer> {
 
                 spec.commandLine().getOut().println("Specified class isn't a valid connector type. The following connector type(s) are available:");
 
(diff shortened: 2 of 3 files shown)


C0urante:
CC @gunnarmorling 😃

src/test/java/org/kcctl/command/ApplyCommandTest.java
```diff
@@ -140,4 +140,14 @@ public void fail_on_first_error() {
...
     }
 
+    @Test
+    public void should_gracefully_handle_missing_connector_class() {
+        var path = Paths.get("src", "test", "resources", "nonexistent.json");
+
+        int exitCode = context.commandLine().execute("-f", path.toAbsolutePath().toString());
+
+        assertThat(exitCode).isEqualTo(CommandLine.ExitCode.SOFTWARE);
+        assertThat(context.output().toString()).contains("Specified class isn't a valid connector type");
```
C0urante, line 150: I wanted to add an assertion about it actually showing the list of available connectors, but was having some trouble running the integration tests locally (containers were hanging on startup).

While less powerful, this test still guards against a regression that would re-introduce the NPE addressed by this PR.

gunnarmorling: Is that container issue on Mac M1? If so, I've recently pushed an update which seems to help for the Kafka image, but Connect still seems slow. It's not quite clear to my why, as the Debezium 2.0 images have an Arm flaviour. Perhaps Testcontainers is not pulling that in for some reason?

gunnarmorling:
Applied; thank you so much, @C0urante!

src/test/java/org/kcctl/command/ApplyCommandTest.java
C0urante: It is indeed on M1!

I've taken another stab at this on the latest main and am no longer seeing these issues. I thought I'd pulled the latest main before creating the branch for this PR, but maybe not?

Either way, filed https://github.com/kcctl/kcctl/pull/302 to add this missing test coverage and (of course 🙃) fix a bug that it unearthed.

gunnarmorling: > It is indeed on M1!
>
> I've taken another stab at this on the latest main and am no longer seeing these issues.

Happy to hear! One thing we could do to further speed things up a bit is to use the _debezium-base_ image rather than _debezium_, as scanning the plug-in path definitely takes some time. That'd require to adjust a few tests though, e.g. on expected connector plug-ins.

gunnarmorling: Logged https://github.com/kcctl/kcctl/issues/303 for this.

PR #302, opened 2022-12-28
#295 (follow-up) Fix configuration context for listing plugins when apply fails

C0urante:
Follow-up that fixes a small bug in https://github.com/kcctl/kcctl/pull/299 and adds better testing coverage for the list-connectors-when-class-not-found logic in `apply`.


gunnarmorling (approved):
LGTM, merging. Thanks, @C0urante!

Issue #2, opened 2021-07-11
Display task offsets


gunnarmorling:
Upon _kcctl describe my-connector_, the task offsets should be shown. This will require to expand the notion of the configuration context, also allowing for the specification of Kafka bootstrap server URLs and the name of the offset topic.

sfc-gh-japatel:
This is how I do it on my machine. 
```
function listConsumerGroup() {
        ~/Desktop/Kafka-Connector/snowflake-kafka-connector/test/confluent-5.5.0/bin/kafka-consumer-groups --bootstrap-server localhost:9092 --list
}

function describeConsumerGroup() {
        ~/Desktop/Kafka-Connector/snowflake-kafka-connector/test/confluent-5.5.0/bin/kafka-consumer-groups --bootstrap-server localhost:9092 --describe --group $1
}
```

I am missing something when you say configuration context.

gunnarmorling:
> I am missing something when you say configuration context.

The idea is that you can use _kcctl_ to work with and switch between multiple "environments", i.e. Kafka Connect instances. Currently, they are just represented by the KC URL, but for the offset display/reset (see #8), we'd also need to add the Kafka bootstrap URL and name of the offset topic. I.e. like so:

```
# Define context named "local"
kcctl config set-context --name local --cluster=http://localhost:8083 --boostrap-servers=my-kafka:9092 --offset-topic=myoffsets

# Switch to another context
kcctl config use pre-prod
```

Any commands subsequently executed, like `kcctl get connectors`, will always use the currently set context.

gunnarmorling:
@helpermethod, in case you wanted to take a stab at offset reset, the change discussed here for expaning the configuration context probably should be the first step. It'll require just a little bit of care to read back the current file format if present, which just contains the KC cluster URL. We might not have to fully support it, but logging an error message in this case would be good: "spotted earlier config file format. Please run kcctl config set-context ... again".

sfc-gh-japatel:
I am happy to take this over if @helpermethod is fine with that. But I dont have any idea on how to achieve this? Is there a rest API which display current offsets? All I know is there is a `kafka-consumer-groups` shell script. 
CC: @gunnarmorling

gunnarmorling:
@sfc-gh-japatel, thanks, you contribution would definitely be welcomed. Let's just wait to hear back from @helpermethod and see whether has made any progress, but I think he's more focused on the side of resetting offsets (#8).

[...]

PR #338, opened 2023-07-26
#2 Add 'get offsets' command

C0urante:
Addresses https://github.com/kcctl/kcctl/issues/2

Usage:

```
Usage: kcctl get offsets [NAME...]
Displays committed connector offsets
      [NAME...]   Name of the connector (e.g. 'my-connector')
```

Uses the `GET /connectors/{name}/offsets` endpoint introduced in [KIP-875](https://cwiki.apache.org/confluence/display/KAFKA/KIP-875%3A+First-class+offsets+support+in+Kafka+Connect#KIP875:FirstclassoffsetssupportinKafkaConnect-Endpoints) to retrieve committed offsets for a connector.

Since this feature requires Kafka Connect 3.5 or later, the integration test is customized to use a non-stable version of Debezium (2.4.0 alpha). We can and should update the version used in that test once a final release takes place for 2.4.0.


src/main/java/org/kcctl/command/GetOffsetsCommand.java
```diff
@@ -0,0 +1,88 @@
+/*
+ *  Copyright 2021 The original authors
```
gunnarmorling, line 2: Logged https://github.com/kcctl/kcctl/issues/340 for getting rid of the year there and use a simpler heade altogether.

src/test/java/org/kcctl/command/GetOffsetsCommandTest.java
```diff
@@ -0,0 +1,130 @@
...
+class GetOffsetsCommandTest extends IntegrationTest {
+
+    private final ObjectMapper mapper = new ObjectMapper();
+
+    @InjectCommandContext
+    KcctlCommandContext<GetOffsetsCommand> context;
+
+    @BeforeAll
+    public static void prepare() {
+        // TODO: replace explicit container tag with DebeziumContainer.latestStable() once Debezium releases 2.4, which is based on Kafka 3.5
```
gunnarmorling, line 55: Could you log a follow-up issue so we don't forget about it?

src/test/java/org/kcctl/command/GetOffsetsCommandTest.java
```diff
@@ -0,0 +1,130 @@
...
+        IntegrationTest.prepare();
+    }
+
+    @Test
+    public void should_list_offsets_single_connector() throws Exception {
+        registerTestConnector("test1");
+
+        Instant start = Instant.now();
+
+        while (Duration.between(start, Instant.now()).toSeconds() < 30) {
```
gunnarmorling, line 71: Looks like a good use case for Awaitility. Let's add it as a dependency and use it here if you like the idea. It would make that test a bit nicer to read, but it's also not a strict requirement from my PoV.

C0urante:
@gunnarmorling if you have a moment, mind taking a look at this one? Thanks!

gunnarmorling (commented):
Woot, woot, LOVING it 😍! This has been on the roadmap for sooo long (I mean, it's issue #2 after all :). Thanks a ton for picking this up.

Looks great to me overall, a few minor comments inline. Besides those, taking a step back, instead of making this a new sub-command of _get_, did you consider making this part of _describe connector_, e.g. via a new switch akin to _--tasks-config_? I feel this would be somewhat more intuitive/user-friendly, as I wouldn't consider offsets to be their own entity that much as just a property of the connector entity. Curious what you think.

Once we've made a decision on  that one,

* The completion file should be updated as part of this PR, too
* The help output in the README should be updated if it changes at all (if we stick with _get offsets_, then the help output of _get_ should mention that)

Thanks again!

C0urante:
Thanks Gunnar! Planning to push a new commit tomorrow to address the suggestions.

RE folding this into `describe connector`--this is really interesting, and I see the argument for intuitiveness. One question that comes to mind is if that approach will remain as intuitive when we add support for altering/resetting connector offsets. I think for altering offsets we could possibly fold it into `patch connector` via something like `--offsets` (though this feels like it might be giving that command a little too much to do), but for resetting them entirely, I'm not sure what we could do... `delete connector --offsets` doesn't seem great, `patch connector --reset-offsets` seems better, but then we have `describe connector --offsets`, `patch connector --offsets`, and `patch connector --reset-offsets` vs. `get offsets`, `patch offsets`, and `delete offsets`, which doesn't seem inherently superior IMO.

Thoughts on which syntax to choose with the future offsets-related commands (that I \*promise\* to implement after Kafka 3.6.0 is released) in mind?

gunnarmorling:
Good point about those future offset related APIs. Looks like making `offsets` their own first-level entity as you propose is more practical indeed.

We still may consider later on whether to add a flag `--offsets` to `describe connector`, so as to get everything connector-related at once (and I think examining offsets will be a more common operation than changing or deleting them). It'd be a bit redundant, though.

I'd say let's go with your proposal and take it from there.

src/test/java/org/kcctl/command/GetOffsetsCommandTest.java
C0urante: 👍 Done: https://github.com/kcctl/kcctl/issues/346

src/test/java/org/kcctl/command/GetOffsetsCommandTest.java
C0urante: Love it, done. The one small issue I found is that (with the five minutes of docs perusal I did) there doesn't seem to be an easy way to emit a custom error message if an awaited condition never evaluates to true, but it still seems worth the tradeoff in improved readability, especially since (fingers crossed) it's unlikely these tests will start spontaneously failing.

C0urante:
Thanks Gunnar! I've updated the README and completion script, and addressed your other inline code suggestions. This is ready for another pass when you have time.

C0urante:
Hmm... the `RestartConnectorCommandTest::should_restart_only_once_with_regexp` test case failed on the latest CI build. The command invocation completed with a non-zero exit status. I wonder if this is because we don't wait for connectors to be up and running before restarting them? If we issue a restart request too soon after creating one or more connectors, it may get rejected by the Connect cluster if a rebalance is about to happen.

Kicking off another CI run. If this is only slightly flaky, I'd like to file a follow-up issue to investigate further; if it occurs frequently enough to be annoying, then we should probably investigate further before merging this PR to make sure that it's not caused by these changes.

C0urante:
Latest run succeeded. Kicking off one more just to be safe.

gunnarmorling:
Yeah, we definitely have some issues with flaky tests. Last one just passed, will merge.

>  If this is only slightly flaky, I'd like to file a follow-up issue to investigate further

Yes, if you could log one, that'd be great. Not sure when this started, also saw this when merging a few Dependabot PRs yesterday. So it definitely shouldn't be related by this PR.

Thanks again, really cool to see this finally becoming reality in kcctl!

C0urante:
Thanks Gunnar! Logged https://github.com/kcctl/kcctl/issues/353 regarding the flaky integration tests.

Issue #339, opened 2023-07-26
Add support for stopping (not deleting) connectors


C0urante:
[KIP-875](https://cwiki.apache.org/confluence/display/KAFKA/KIP-875%3A+First-class+offsets+support+in+Kafka+Connect#KIP875:FirstclassoffsetssupportinKafkaConnect-Newtargetstate:STOPPED) introduced a new target state for connectors: `STOPPED`. This state differs from `PAUSED` in that it completely shuts down all tasks for the connector, instead of leaving them active but on standby.

It's also a requirement for connectors to be in that state before their offsets can be [reset](https://cwiki.apache.org/confluence/display/KAFKA/KIP-875%3A+First-class+offsets+support+in+Kafka+Connect#KIP875:FirstclassoffsetssupportinKafkaConnect-Resettingoffsets) or [altered](https://cwiki.apache.org/confluence/display/KAFKA/KIP-875%3A+First-class+offsets+support+in+Kafka+Connect#KIP875:FirstclassoffsetssupportinKafkaConnect-Alteringoffsets).

Since support for this new state was added in Kafka 3.5, we can and should add support for it in kcctl.

gunnarmorling:
Definitely makes sense to support this. PR welcome :)

C0urante:
Self-assigned, plan on implementing tomorrow :)

gunnarmorling:
Awesome, thanks a lot!

Issue #346, opened 2023-07-27
Remove unstable Docker testing image for offset read and stop connector tests


C0urante:
Created after discussion on https://github.com/kcctl/kcctl/pull/338

We use the latest stable tag for the Debezium Kafka Connect image in our integration tests. Currently, that tag is [2.3.0.Final](https://hub.docker.com/layers/debezium/connect/2.3.0.Final/images/sha256-aebae17f1971aab6629a1a7994d7de95bc5befe43baaa2173080275692599432), which is built with Kafka Connect 3.4. That version of Kafka Connect doesn't support the `GET /connectors/{name}/offsets` or `PUT /connectors/{name}/stop` endpoints, so we use a different, unstable tag in the tests for the `get offsets` and `stop connector` commands that corresponds to an image built with Kafka Connect 3.5.

Once Debezium publishes a stable image built off of Kafka Connect 3.5, we can and should remove that override from those tests.

C0urante:
Nope, believe we took care of it when we bumped everything to a `.Final` tag!

C0urante:
Plus, for any future tests, we can use the trunk build option instead of relying on unstable Debezium tags (although it's probably fine if we have to resort to that for one reason or another).

PR #347, opened 2023-07-27
#339 Add 'stop connector' command

C0urante:
Addresses https://github.com/kcctl/kcctl/issues/339

Usage:

```
Usage: kcctl stop connector [-e] [NAME...]
Stops (but does not delete) the specified connector(s)
      [NAME...]   Name(s) of the connector(s) (e.g. 'my-connector').
  -e, --reg-exp   use NAME(s) as regexp pattern(s) to use on all connectors
```

Uses the `PUT /connectors/{name}/stop` endpoint introduced in [KIP-875](https://cwiki.apache.org/confluence/display/KAFKA/KIP-875%3A+First-class+offsets+support+in+Kafka+Connect#KIP875:FirstclassoffsetssupportinKafkaConnect-Endpoints) to stop (but not delete) connectors.

Similar to the 'get offsets' command, this feature requires Kafka Connect 3.5 or later, and the integration test is customized to use a non-stable version of Debezium (2.4.0 alpha). We can and should update the version used in that test once a final release takes place for 2.4.0.

### DIFF
--- a/src/test/java/org/kcctl/IntegrationTest.java
+++ b/src/test/java/org/kcctl/IntegrationTest.java
@@ -54,11 +54,28 @@ public abstract class IntegrationTest {
     protected static final KafkaContainer kafka = new KafkaContainer(DockerImageName.parse("confluentinc/cp-kafka:7.3.1"))
             .withNetwork(network);
 
-    protected static final DebeziumContainer kafkaConnectBase = DebeziumContainer.latestStable()
+    protected static final DebeziumContainer kafkaConnectLatestStable = DebeziumContainer.latestStable()
             .withNetwork(network)
             .withKafka(kafka)
-            .dependsOn(kafka);
-    protected static DebeziumContainer kafkaConnect = kafkaConnectBase;
+            .dependsOn(kafka)
+            .withEnv("GROUP_ID", "debezium_latest_stable");
+
+    // TODO: remove this separate container once Debezium releases 2.4, which is based on Kafka 3.5;
+    // we can just go back to kafkaConnectLatestStable for all of our tests then
+    protected static final DebeziumContainer kafkaConnect24Alpha = new DebeziumContainer("debezium/connect:2.4.0.Alpha1")
+            .withNetwork(network)
+            .withKafka(kafka)
+            .dependsOn(kafka)
+            // Force rapid offset commits, in order for offsets to appear quickly in the REST API for the 'get offsets' command tests
+            .withEnv("OFFSET_FLUSH_INTERVAL_MS", "1000")
+            // Override the group ID to prevent this worker from forming a cluster with workers in other containers
+            .withEnv("GROUP_ID", "debezium_2_4_alpha")
+            // Can't share internal topics, either
+            .withEnv("CONFIG_STORAGE_TOPIC", "debezium_2_4_alpha_configs")
+            .withEnv("OFFSET_STORAGE_TOPIC", "debezium_2_4_alpha_offsets")
+            .withEnv("STATUS_STORAGE_TOPIC", "debezium_2_4_alpha_status");
+
+    protected static DebeziumContainer kafkaConnect = kafkaConnectLatestStable;
 
     @BeforeAll
     public static void prepare() {
@@ -93,7 +110,14 @@ public abstract class IntegrationTest {
     @AfterAll
     public static void resetKafkaConnectImage() {
         // In case a test overrode the Kafka Connect image
-        kafkaConnect = kafkaConnectBase;
+        kafkaConnect = kafkaConnectLatestStable;
+    }
+
+    protected void registerTestConnectors(String name, String... names) {
+        registerTestConnector(name);
+        for (String extraName : names) {
+            registerTestConnector(extraName);
+        }
     }
 
     protected void registerTestConnector(String name) {
--- a/src/test/java/org/kcctl/command/StopConnectorCommandTest.java
+++ b/src/test/java/org/kcctl/command/StopConnectorCommandTest.java
@@ -1,0 +1,162 @@
+/*
+ *  Copyright 2021 The original authors
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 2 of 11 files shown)


src/test/java/org/kcctl/IntegrationTest.java
```diff
@@ -88,6 +90,19 @@ public void cleanup() {
         kafkaConnect.deleteAllConnectors();
     }
 
+    @AfterAll
+    public static void resetKafkaConnectImage() {
+        // In case a test overrode the Kafka Connect image
+        kafkaConnect = kafkaConnectBase;
+    }
+
+    protected void registerTestConnectors(String name, String... names) {
```
C0urante, line 99: If this approach looks good, I can file a follow-up to adapt it wherever applicable across the rest of our integration tests.

src/test/java/org/kcctl/command/StopConnectorCommandTest.java
```diff
@@ -0,0 +1,170 @@
...
+        for (String connector : connectors) {
+            // Make sure the connector and one task have actually started
+            kafkaConnect.ensureConnectorState(connector, Connector.State.RUNNING);
+            kafkaConnect.ensureConnectorTaskState(connector, 0, Connector.State.RUNNING);
+        }
+    }
+
+    // TODO: Replace this with kafkaConnect.ensureConnectorState if/when
+    //       the Debezium testing library is updated with support for the STOPPED state
+    //       (https://github.com/debezium/debezium/pull/4709)
```
C0urante, line 144: Can file an issue to track this item if this temporary workaround looks good.

src/test/java/org/kcctl/IntegrationTest.java
gunnarmorling: Yeah!

gunnarmorling (approved):
LGTM, thanks a lot! +1 for logging all the discussed follow-up issues. And this will need a rebase due to the merge conflicts with main now, after having merged the offsets PR.

C0urante:
Turns out the test failures in this one weren't flaky! The worker in the 2.4 Debezium image was automatically forming a cluster with the one in the latest stable (i.e., 2.3) image, which also tended to be the leader whenever the full set of tests was run but not when the `StopConnectorCommandTest` suite was run in isolation. Bit of a headscratcher to debug but I think the latest fix I've pushed should work. The reason this wasn't an issue with https://github.com/kcctl/kcctl/pull/338 is that the `GET /connector/{name}/offsets` endpoint doesn't require forwarding to the leader, whereas the `PUT /connector/{name}/stop` endpoint does.

src/test/java/org/kcctl/IntegrationTest.java
```diff
@@ -54,11 +56,28 @@ public abstract class IntegrationTest {
...
+            .withKafka(kafka)
+            .dependsOn(kafka)
+            // Force rapid offset commits, in order for offsets to appear quickly in the REST API for the 'get offsets' command tests
+            .withEnv("OFFSET_FLUSH_INTERVAL_MS", "1000")
+            // Override the group ID to prevent this worker from forming a cluster with workers in other containers
+            .withEnv("GROUP_ID", "debezium_2_4_alpha")
+            // Can't share internal topics, either
+            .withEnv("CONFIG_STORAGE_TOPIC", "debezium_2_4_alpha_configs")
+            .withEnv("OFFSET_STORAGE_TOPIC", "debezium_2_4_alpha_offsets")
+            .withEnv("STATUS_STORAGE_TOPIC", "debezium_2_4_alpha_status");
```
C0urante, line 78: I created a single reusable 2.4 image for testing both the `get offsets` and `stop connector` commands. Should reduce runtime bloat and help us track which tests are using unstable images when it comes time to address https://github.com/kcctl/kcctl/issues/346.

src/test/java/org/kcctl/IntegrationTest.java
C0urante: Cool! As soon as this is merged I'll throw up a PR.

src/test/java/org/kcctl/IntegrationTest.java
```diff
@@ -93,7 +112,28 @@ public void cleanup() {
...
-        kafkaConnect = kafkaConnectBase;
+        kafkaConnect = kafkaConnectLatestStable;
+    }
+
+    /**
+     * Run a command with the given arguments and verify that it has exited successfully
+     * @param context the {@link KcctlCommandContext} for the command
+     * @param args the arguments with which the command should be invoked
+     */
+    protected static void runAndCheckExitCode(KcctlCommandContext<?> context, String... args) {
```
C0urante, line 123: This was helpful when debugging the test failures which were only cropping up when multiple integration test suites were run (which made it difficult to run things through a debugger in my IDE).

Similar to the varargs variant of `registerTestConnectors`, if this looks good, can adapt the rest of the integration test code base to use it wherever applicable.

C0urante:
@gunnarmorling thanks again for all the rapid reviews! LMK if you'd like to take another look; otherwise, I can merge and we can see about putting out a new release that includes both this and the `get offsets` feature.

src/test/java/org/kcctl/IntegrationTest.java
C0urante: Tweaked this method to be a public instance method for `KcctlCommandContext` instead of a protected static method for `IntegrationTest`, but the general idea is the same.

src/test/java/org/kcctl/IntegrationTest.java
C0urante: Just kidding, published https://github.com/kcctl/kcctl/pull/354

src/test/java/org/kcctl/IntegrationTest.java
C0urante: Applied this change across the board in https://github.com/kcctl/kcctl/pull/354

gunnarmorling:
> The worker in the 2.4 Debezium image was automatically forming a cluster with the one in the latest stable (i.e., 2.3) image

Oh, wow, tricky one indeed.

gunnarmorling:
LGTM, merging. Thanks a lot, @C0urante!

gunnarmorling:
> Addresses https://github.com/kcctl/kcctl/issues/339

Btw. if you say "Fixes https://github.com/kcctl/kcctl/issues/339", the issue will be automatically closed when merging the PR (just manually did it).

PR #354, opened 2023-07-28
MINOR: Sprinkle some syntactic sugar on our integration tests

C0urante:
- Replaces consecutive calls to `IntegrationTest::registerTestConnector` with single calls to `IntegrationTest::registerTestConnectors`
- Replaces all manual verifications that a command exited with status zero with `KcctlCommandContext::runAndCheckExitCode`, which on top of asserting that the command exited successfully, provides the complete contents of stdout/stderr if that assertion fails

This has unearthed a non-deterministic NPE that's occurring in the `PatchConnectorCommandTest` because the `DescribeConnectorCommand` instance that's created and used by `PatchConnectorCommand` has a null `spec` field. Strangely, this is not occurring consistently on my machine. I've been able to reproduce it as far back as 57afec337aecd3ab7cc2bbb65782b682fa51f3bb, so at the very least it's not caused by this PR, or the recent PRs to implement the `get offsets` and `stop connector` commands. We can probably investigate as a follow-up item, and that may even be easier with the changes here (🤞).

### DIFF
--- a/src/test/java/org/kcctl/support/KcctlCommandContext.java
+++ b/src/test/java/org/kcctl/support/KcctlCommandContext.java
@@ -37,7 +37,7 @@ public record KcctlCommandContext<T>(T command, CommandLine commandLine, StringW
      * Run a command with the given arguments and verify that it has exited successfully
      * @param args the arguments with which the command should be invoked
      */
-    public void runAndCheckExitCode(String... args) {
+    public void runAndEnsureExitCodeOk(String... args) {
         int exitCode = commandLine.execute(args);
         assertThat(exitCode)
                 .overridingErrorMessage(() -> "Command with args '" + String.join("', '", args) + "' "
--- a/src/test/java/org/kcctl/IntegrationTest.java
+++ b/src/test/java/org/kcctl/IntegrationTest.java
@@ -62,6 +62,7 @@ public abstract class IntegrationTest {
 
     // TODO: remove this separate container once Debezium releases 2.4, which is based on Kafka 3.5;
     // we can just go back to kafkaConnectLatestStable for all of our tests then
+    // (https://github.com/kcctl/kcctl/issues/346)
     protected static final DebeziumContainer kafkaConnect24Alpha = new DebeziumContainer("debezium/connect:2.4.0.Alpha1")
             .withNetwork(network)
             .withKafka(kafka)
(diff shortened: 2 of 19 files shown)


src/test/java/org/kcctl/support/KcctlCommandContext.java
```diff
@@ -30,4 +32,19 @@ public void reset() {
         output.getBuffer().setLength(0);
         error.getBuffer().setLength(0);
     }
+
+    /**
+     * Run a command with the given arguments and verify that it has exited successfully
+     * @param args the arguments with which the command should be invoked
+     */
+    public void runAndCheckExitCode(String... args) {
```
gunnarmorling, line 40: "check exit code" to me suggests that I could specify which code I'm expecting. But given we expect that the command succeeds, making this more explicit in the name would be nice. How about `runAndEnsureExitCodeOk()`, `runAndEnsureStatusOk()`, `runAndEnsureSuccessfulExecution()`, or something along those lines?

gunnarmorling (commented):
Very nice. One small naming suggestion inline.

gunnarmorling:
Btw. the test flakiness situation seems to have deteriorated recently; most PR builds seem to fail atm., which didn't use to be the case. But from a quick look, there's a range of issues, not just one which would explain it 🤔 .

gunnarmorling:
@C0urante, could you rebase this one, also amending any remaining TODOs with issue references, where needed? Thx!

src/test/java/org/kcctl/support/KcctlCommandContext.java
C0urante: Thanks, I agree that it's a little unclear. Went with `runAndEnsureExitCodeOk`.

C0urante:
Think I figured out the root cause for the integration test flakiness. Opened https://github.com/kcctl/kcctl/pull/356 to address, will try to get in a few consecutive CI runs to validate.

src/test/java/org/kcctl/support/KcctlCommandContext.java
gunnarmorling: Perfect, thx!

gunnarmorling (approved):
LGTM, merging. Thanks, @C0urante!

PR #356, opened 2023-07-28
#353 Fix flaky integration tests

### LINKED ISSUE #353: Flaky integration tests
Some of our integration tests are a little flaky and have been causing CI builds to fail. See builds on https://github.com/kcctl/kcctl/pull/338 for some examples.

It's possible that some of this flakiness is caused by creating connectors and then immediately hitting the Connect cluster again with a request that fails because a rebalance is pending (rebalances are usually triggered twice when connectors are created: once to assign the `Connector` instance to a worker, and a second time to assign its `Task` instances to workers across the cluster).

C0urante:
Fixes #353, depends on #354

The `DescribeConnectorCommand` is manually invoked by the `PatchConnectorCommand`, but without being given to a wrapping `CommandLine` instance. As a result, its `@CommandLine.Spec`-annotated `spec` field is not initialized in that case, and under some rare circumstances, the command will try to emit output via `spec.commandLine().getOut()::print` (or `println`, or `printf`, etc.). This causes a `NullPointerException`.

The scenario that leads to the attempted output is when the Connect cluster serves a 409 response due to a pending rebalance. This is otherwise gracefully handled with bounded retry logic by the `PatchConnectorCommand`.

This fix is a bit quick-and-dirty in that it doesn't prevent the `NullPointerException` from occurring in the extremely-rare event that the connector is not found, but that situation should never arise except from rapid conflicting concurrent operations on the Connect cluster. We can and should investigate a more fully-fledged fix, but this is one quick and lightweight option that may suffice for now to patch the recently-occurring integration test flakiness.

### DIFF
--- a/src/main/java/org/kcctl/command/DescribeConnectorCommand.java
+++ b/src/main/java/org/kcctl/command/DescribeConnectorCommand.java
@@ -26,6 +26,7 @@ import org.kcctl.completion.ConnectorNameCompletions;
 import org.kcctl.service.ConnectorInfo;
 import org.kcctl.service.ConnectorStatusInfo;
 import org.kcctl.service.KafkaConnectApi;
+import org.kcctl.service.KafkaConnectException;
 import org.kcctl.service.TaskState;
 import org.kcctl.service.TopicsInfo;
 import org.kcctl.util.Colors;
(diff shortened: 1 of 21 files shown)


C0urante:
Run 1 is green; kicking off another.

C0urante:
Run 2 is green; kicking off another.

C0urante:
Run 3 is green; kicking off another.

C0urante:
Run 4 is green; kicking off just one more.

gunnarmorling (approved):
Woohoo, great analysis, and really happy to see this is fixed. Please feel free to log a follow-up issue, if you think further investigation is needed to handle that error situation more holistically. Thanks a lot!

PR #359, opened 2023-07-29
#357 Retry all REST requests on 409 errors

### LINKED ISSUE #357: Make PatchConnectorCommand more robust againt concurrent config changes
See https://github.com/kcctl/kcctl/actions/runs/5699964560/job/15449368506; the scope of the retry loop around one of the API calls seems not to be enough, we should do something similar for the other (read) API call. Perhaps we can extract this retry loop into a re-usable method like so:

```
Map<String, String> connectorParameters = retrying(() -> kafkaConnectApi.getConnectorConfig(conectorToPatch));
```

C0urante:
Fixes #357

This uses a different approach than what was proposed in the issue, but the effect should be mostly the same. I drafted up the `retrying(() -> kafkaConnectApi.getConnectorConfig(conectorToPatch))` patch but realized that this would still have to be manually applied to all REST calls that could result in a 409 response, which seemed fairly brittle.

I did some research and it looks like we can get retry logic with our auto-generated REST client for free by bringing in a small extra dependency: https://quarkus.io/guides/smallrye-fault-tolerance

Since the API for describing retriable requests with this library is based on exception class, I added custom subclasses to the `KafkaConnectException` to help us distinguish between retriable and non-retriable operations. We could theoretically create a `RetriableKafkaConnectException` umbrella class for those, but that feels a little premature, especially since some exceptions may be retriable for only a subset of requests that throw them.

I was able to get 10 consecutive green integration test runs locally with this patch.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -86,6 +86,10 @@
       <artifactId>quarkus-rest-client</artifactId>
     </dependency>
     <dependency>
+      <groupId>io.quarkus</groupId>
+      <artifactId>quarkus-smallrye-fault-tolerance</artifactId>
+    </dependency>
+    <dependency>
       <groupId>com.github.freva</groupId>
       <artifactId>ascii-table</artifactId>
       <version>1.8.0</version>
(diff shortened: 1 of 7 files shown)


gunnarmorling:
Oooh, that's lovely! Much more elegant indeed, great find, and great improvement 🥳 !

PR #360, opened 2023-07-30
#355 Matrix testing on CI

### LINKED ISSUE #355: Set up matrix testing on CI
We've started to have a mix of Debezium container image / Connect versions in the test suite, so as to support recently added Connect API calls. This created some challenges when the workers of different KC versions formed a cluster when keeping the Docker container(s) running for the duration of the entire test suite.

We could avoid this kind of issue, and also make sure any "emulated" kcctl features (where we fall back to multiple API calls on older Connect versions) are tested thoroughly, by setting up matrix testing on CI. We would:

* Work against only a single Connect version in the test code (defaulting to the latest one we support)
* Add an annotation like `@SkipIfConnectVersionIsOlderThan("3.5")` for excluding tests for new features when testing against older versions
* Make the version to test against configurable via an env var
* Configure a matrix test job on GH Actions for testing against all the versions we want to support (not sure what would make sense, perhaps all 3.x minors and latest 2.x minor? Or is this too much?)

@C0urante, curious what you think?

C0urante:
(Draft opened solely for testing, not intended for review or merge yet)

Fixes #355 

Runs integration tests against a wide range of versions of Kafka Connect using the approach outlined in the linked issue.


.github/workflows/maven.yml
```diff
@@ -51,3 +51,20 @@ jobs:
 
       - name: 'Build'
         run: ./mvnw -ntp -B --file pom.xml verify
+        env:
+          DEBEZIUM_TAG: ${{ matrix.debezium_tag }}
```
gunnarmorling, line 55: Instead of passing in the Debezium tag, could we make this to use the Kafka version (as that's what matters in the end for controlling which tests to run). E.g. when a test fails due to an unsupported API, we'd easily say that this is about, say, Kafka Connect 3.2, rather than some largely unrelated Debezium version. This would require to codify the Kafka-to-Debezium version mapping somewhere in the code which spins up Testcontainers.

gunnarmorling (commented):
This looks great! Two minor comments/suggestions inline.

src/test/java/org/kcctl/command/GetPluginsCommandTest.java
```diff
@@ -37,33 +40,61 @@
...
+        String debeziumVersion = getDebeziumVersion();
         String kafkaVersion = getConnectVersion();
 
+        Version parsedDebeziumVersion = new Version(debeziumVersion);
+        Version parsedKafkaVersion = new Version(kafkaVersion);
+
+        String spannerConnectorDescription = null;
+        if (parsedDebeziumVersion.greaterOrEquals(SPANNER_CONNECTOR_MIN_VERSION)) {
+            spannerConnectorDescription = "source   io.debezium.connector.spanner.SpannerConnector              " + debeziumVersion;
+        }
```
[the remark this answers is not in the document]
C0urante: +1 for switching to `connect-base`, although we'd still end up doing most of the dancing here since the rest (besides the Spanner connector) concerns MM2 and the file connectors :)

.github/workflows/maven.yml
C0urante: Good call, done! I've left in the latest stable and nightly images since they may help us catch breaking changes with Connect sooner, but can remove those too if they don't seem worth the effort.

src/main/java/org/kcctl/service/KafkaConnectException.java
```diff
@@ -28,11 +28,16 @@ public static KafkaConnectException from(Response response) {
 
         ErrorResponse errorResponse = response.readEntity(ErrorResponse.class);
 
-        return switch (response.getStatusInfo().toEnum()) {
-            case CONFLICT -> new KafkaConnectConflictException(errorResponse.message);
-            case NOT_FOUND -> new KafkaConnectNotFoundException(errorResponse.message);
-            default -> new KafkaConnectException(errorResponse);
-        };
+        if (response.getStatusInfo() == Response.Status.CONFLICT
+                || errorResponse.message.contains("Request cannot be completed because a rebalance is expected")) {
```
C0urante, line 32: Ugh... turns out there's an ancient bug in Connect where the "rebalance pending" response can have either a 409 status code (correct) or a 500 status code (incorrect). Best we can do for now is go back to manually examining the error message to handle that case.

We could technically change this behavior in Connect, and it is probably worth addressing, but for tools like this one that strive for backwards compatibility with a wide range of versions, that won't buy us much for major/minor versions that are already EOL.

src/test/java/org/kcctl/command/GetConnectorsCommandTest.java
```diff
@@ -49,6 +51,10 @@ public void should_print_empty_connectors_() {
     @Test
     public void should_print_registered_connectors() {
         registerTestConnectors("test1", "test2");
+        for (String connector : List.of("test1", "test2")) {
+            kafkaConnect.ensureConnectorState(connector, Connector.State.RUNNING);
+            kafkaConnect.ensureConnectorTaskState(connector, 0, Connector.State.RUNNING);
```
C0urante, line 56: This test failed in a [prior CI run](https://github.com/kcctl/kcctl/actions/runs/5705038378/job/15459252836). In rare cases, we'd successfully submit the connector to the Connect cluster, and then before tasks for it could be brought up, run the `get connectors` command.

The calls to `ensureConnectorState` and (more importantly) `ensureConnectorTaskState` here ensure that the connector and its task have actually started before we examine the cluster using the CLI.

src/test/java/org/kcctl/command/GetPluginsCommandTest.java
```diff
@@ -37,33 +40,61 @@
...
-                        "source   io.debezium.connector.sqlserver.SqlServerConnector          " + debeziumVersion,
-                        "source   io.debezium.connector.vitess.VitessConnector                " + debeziumVersion,
-                        "source   org.apache.kafka.connect.mirror.MirrorCheckpointConnector   " + kafkaVersion,
-                        "source   org.apache.kafka.connect.mirror.MirrorHeartbeatConnector    " + kafkaVersion,
-                        "source   org.apache.kafka.connect.mirror.MirrorSourceConnector       " + kafkaVersion);
+                .containsExactly(expectedOutput);
     }
 
     @Test
+    @SkipIfConnectVersionIsOlderThan("3.2")
```
C0urante, line 97: This syntax is so smooth and unobtrusive. Great idea, Gunnar!

src/main/java/org/kcctl/service/KafkaConnectException.java
gunnarmorling: Gasp, yeah, reality just sucks sometimes ;) But actually also nice demonstration of the value of setting up this matrix test.

src/test/java/org/kcctl/command/GetConnectorsCommandTest.java
gunnarmorling: Nice find!

src/test/java/org/kcctl/command/GetPluginsCommandTest.java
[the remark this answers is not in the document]
gunnarmorling: LOL, yeah, fair point. The other advantage would be faster test execution, as the plug-in path scanning in Connect feels surprisingly slow.

src/test/java/org/kcctl/command/GetPluginsCommandTest.java
gunnarmorling: Ha yeah, thanks a lot for setting this up! We have something similar in Debezium, worked great there as well.

.github/workflows/maven.yml
```diff
@@ -51,3 +51,20 @@ jobs:
 
       - name: 'Build'
         run: ./mvnw -ntp -B --file pom.xml verify
+        env:
+          CONNECT_VERSION: ${{ matrix.connect_version }}
+
+    strategy:
+      fail-fast: false
```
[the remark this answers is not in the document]
gunnarmorling: Gotcha!

gunnarmorling (approved):
_Great_ improvement, thanks a lot! LGTM, merging.

src/test/java/org/kcctl/command/GetPluginsCommandTest.java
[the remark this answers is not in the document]
gunnarmorling: Noice!

PR #362, opened 2023-07-31
#102 Support new task restart semantics

### LINKED ISSUE #102: Support new restart semantics in Kafka 3.0
Kafka Connect 3.0 has the ability to restart all the failed tasks of a connector:

    POST /connectors/{name}/restart?includeTasks=<true|false>&onlyFailed=<true|false>

More info:

* https://cwiki.apache.org/confluence/display/KAFKA/KIP-745%3A+Connect+API+to+restart+connector+and+tasks
* https://github.com/apache/kafka/pull/10975

The existing `kcctl restart connector <name>` command should be enhanced with a parameter for leveraging this: `--tasks=<all|failed>`. ~~Using these parameters with Connect < 3.0 should be ignored but log a warning. Completion of those options should not be done for Connect 3.0.~~

Thinking more about it, we actually could support this for any KC version, using the new (and more efficient API) if available, otherwise falling back to multiple separate API calls.

C0urante:
Fixes #102

Usage:
```
Usage: kcctl restart connector [-e] [-t=<tasks>] [NAME...]
Restarts the specified connector(s)
      [NAME...]         Name of the connector (e.g. 'my-connector')
  -e, --reg-exp         use NAME(s) as regexp pattern(s) to use on all
                          connectors
  -t, --tasks=<tasks>   also restart tasks for the connector(s); valid values:
                          all, failed
```

I was toying with the idea of not requiring an argument to the `--tasks` flag and defaulting to `all` in that case, but it seemed a little too complex for users. Still, `kcctl restart connector -t file-source` does feel noticeably less clunky than `kcctl restart connector -t all file-source`, so it'd be nice to hear other perspectives on whether that'd be worth supporting.


src/main/java/org/kcctl/service/KafkaConnectApi.java
```diff
@@ -71,6 +71,13 @@ public interface KafkaConnectApi {
     @Path("/connectors/{name}/restart")
     void restartConnector(@PathParam("name") String name);
 
+    @POST
+    @Path("/connectors/{name}/restart")
+    ConnectorStatusInfo restartConnectorAndTasks(
+                                                 @PathParam("name") String name,
+                                                 @QueryParam("includeTasks") boolean includeTasks,
+                                                 @QueryParam("onlyFailed") boolean onlyFailed);
```
C0urante, line 79: I really wanted to omit the `includeTasks` parameter here and hardcode it to be `true` in the REST request, but couldn't find a way to do that. Based on https://github.com/eclipse/microprofile-rest-client/issues/315 it may not even be possible.

gunnarmorling:
> I was toying with the idea of not requiring an argument

Good question. I don't think it's a huge burden for users to specify a value for `-t`, so I'd go with that. We can still make it optional and add a default later on should we feel that's truly better; while in contrast, we couldn't change it from optional to mandatory, so I'd start with it being a required option, if that makes sense.

PR #364, opened 2023-07-31
#363 Add nested subcommand support for help

### LINKED ISSUE #363: Show help for nested subcommands
There's no official way to see help messages for nested subcommands.

The usage for `kcctl` is easy to view, via `kcctl help`, `kcctl -h`, or `kcctl --help`, and provides a list of all commands along with a brief description of what each does.

The usage for any top-level commands (e.g., `kcctl apply`) is also easy to view, via `kcctl help <command>` (e.g., `kcctl help apply` or `kcctl help delete`).

If any of these commands contains nested subcommands, the subcommands are listed with a brief description of what they do. For example:

```
> kcctl help delete
Usage: kcctl delete [<args>...] [COMMAND]
Deletes connectors
      [<args>...]
Commands:
  connector  Deletes specified connector
```

However, there's no way to get usage information for those nested subcommands that doesn't involve hacks like throwing in an unrecognized flag (e.g., `-h`) that causes the command to exit with a non-zero status and print an error message. For example:

```
> kcctl delete connector -h
Unknown option: '-h'
Usage: kcctl delete connector [-e] [CONNECTOR NAME...]
Deletes specified connectors
      [CONNECTOR NAME...]   Name of the connector
  -e, --reg-exp             use CONNECTOR NAME(s) as regexp pattern(s) to apply
                              on all connectors
```

We should make it as easy as possible for users to learn about the syntax/semantics of subcommands. Some options for this include:

1. `kcctl help <command> <subcommand>` (right now this only shows the usage information for `<command>`)
2. `kcctl <command> help <subcommand>` (right now this prints an error message and then the usage information for `<command>`)
3. `kcctl <command> <subcommand> -h` and/or `kcctl <command> <subcommand> --help` (right now this prints an error message and then the usage information for `<subcommand>`)
4. `kcctl <command> <subcommand> help` (right now this invokes the subcommand with an argument of `help`)

IMO option 1 has the most intuitive syntax, but may be difficult to implement. Options 2 and 3 are straightforward to implement via the existing `CommandLine.HelpCommand` class and `mixinStandardHelpOptions` annotation option, respectively. Option 4 is also straightforward to implement with the `CommandLine.HelpCommand` class, but is a little uglier as it causes the subcommand to have its own subcommand named `help`, which pollutes usage messages a bit.

I'd like to implement option 3 unconditionally (`-h` is a deeply-ingrained reflex for many CLI users), and then either option 1 (if it can be done with reasonable effort) or option 2 (if option 1 can't be done). Interested in others' perspectives!

C0urante:
Fixes #363 

Augments the `help` command to process nested subcommands, and adds support for the `-h`/`--help` flag to every subcommand.

Before:
```
> kcctl help restart connectors
Usage: kcctl restart [COMMAND]
Restarts some connectors or a task
Commands:
  connector, connectors  Restarts the specified connector(s)
  task                   Restarts the specified connector or task
```

(This shows the usage info for `restart`, even though we requested it for `restart connectors`)

After:
```
> kcctl help restart connectors
Usage: kcctl restart connector [-eh] [-t=<tasks>] [NAME...]
Restarts the specified connector(s)
      [NAME...]         Name of the connector (e.g. 'my-connector')
  -e, --reg-exp         use NAME(s) as regexp pattern(s) to use on all
                          connectors
  -h, --help            Show this help message and exit.
  -t, --tasks=<tasks>   also restart tasks for the connector(s); valid values:
                          all, failed
```

Before:
```
> kcctl restart connectors -h
Unknown option: '-h'
Usage: kcctl restart connector [-e] [NAME...]
Restarts the specified connector(s)
      [NAME...]   Name of the connector (e.g. 'my-connector')
  -e, --reg-exp   use NAME(s) as regexp pattern(s) to use on all connectors
```
(Prints an error message about `-h` being an unknown option)

After:
```
> kcctl restart connectors -h
Usage: kcctl restart connector [-eh] [-t=<tasks>] [NAME...]
Restarts the specified connector(s)
      [NAME...]         Name of the connector (e.g. 'my-connector')
  -e, --reg-exp         use NAME(s) as regexp pattern(s) to use on all
                          connectors
  -h, --help            Show this help message and exit.
  -t, --tasks=<tasks>   also restart tasks for the connector(s); valid values:
                          all, failed
```

There is one small drawback to this approach in that it adds a bit of noise to the usage summary for every command. Commands that accepted a single argument or flag now have info on one additional flag (`-h`/`--help`).

Autocomplete is supported, but only for top-level commands. This behavior mirrors current autocomplete support for the `help` command, which repeatedly recommends top-level commands even though only the first one is used. Ideally we could support autocomplete for nested subcommands, but this may be difficult to do with picocli.

### DIFF
--- a/kcctl_completion
+++ b/kcctl_completion
@@ -297,7 +297,7 @@ function _picocli_kcctl_info() {
   local curr_word=${COMP_WORDS[COMP_CWORD]}
 
   local commands=""
-  local flag_opts=""
+  local flag_opts="-h --help"
   local arg_opts=""
 
   if [[ "${curr_word}" == -* ]]; then
@@ -315,7 +315,7 @@ function _picocli_kcctl_config() {
   local curr_word=${COMP_WORDS[COMP_CWORD]}
 
   local commands="set-context get-contexts current-context use-context"
-  local flag_opts=""
+  local flag_opts="-h --help"
   local arg_opts=""
 
   if [[ "${curr_word}" == -* ]]; then
@@ -333,7 +333,7 @@ function _picocli_kcctl_get() {
   local curr_word=${COMP_WORDS[COMP_CWORD]}
 
   local commands="plugins connectors offsets loggers logger"
-  local flag_opts=""
+  local flag_opts="-h --help"
   local arg_opts=""
 
   if [[ "${curr_word}" == -* ]]; then
@@ -351,7 +351,7 @@ function _picocli_kcctl_describe() {
   local curr_word=${COMP_WORDS[COMP_CWORD]}
 
   local commands="connector connectors plugin"
-  local flag_opts=""
+  local flag_opts="-h --help"
   local arg_opts=""
 
   if [[ "${curr_word}" == -* ]]; then
@@ -370,7 +370,7 @@ function _picocli_kcctl_apply() {
   local prev_word=${COMP_WORDS[COMP_CWORD-1]}
 
   local commands=""
-  local flag_opts="--dry-run"
+  local flag_opts="-h --help --dry-run"
   local arg_opts="-f --file -n --name"
 
   type compopt &>/dev/null && compopt +o default
@@ -399,7 +399,7 @@ function _picocli_kcctl_patch() {
   local curr_word=${COMP_WORDS[COMP_CWORD]}
 
   local commands="logger connector connectors"
-  local flag_opts=""
+  local flag_opts="-h --help"
   local arg_opts=""
 
   if [[ "${curr_word}" == -* ]]; then
@@ -417,7 +417,7 @@ function _picocli_kcctl_restart() {
   local curr_word=${COMP_WORDS[COMP_CWORD]}
 
   local commands="connector connectors task"
-  local flag_opts=""
+  local flag_opts="-h --help"
   local arg_opts=""
 
   if [[ "${curr_word}" == -* ]]; then
@@ -435,7 +435,7 @@ function _picocli_kcctl_pause() {
   local curr_word=${COMP_WORDS[COMP_CWORD]}
 
   local commands="connector connectors"
-  local flag_opts=""
+  local flag_opts="-h --help"
   local arg_opts=""
 
   if [[ "${curr_word}" == -* ]]; then
@@ -453,7 +453,7 @@ function _picocli_kcctl_resume() {
   local curr_word=${COMP_WORDS[COMP_CWORD]}
 
   local commands="connector connectors"
-  local flag_opts=""
+  local flag_opts="-h --help"
   local arg_opts=""
 
   if [[ "${curr_word}" == -* ]]; then
@@ -471,7 +471,7 @@ function _picocli_kcctl_stop() {
   local curr_word=${COMP_WORDS[COMP_CWORD]}
 
   local commands="connector connectors"
-  local flag_opts=""
+  local flag_opts="-h --help"
   local arg_opts=""
 
   if [[ "${curr_word}" == -* ]]; then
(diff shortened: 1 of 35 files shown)


gunnarmorling:
> (This shows the usage info for restart, even though we requested it for restart connectors)

I'm confused by this one, it still shows the wrong one, no?

```
> kcctl help restart connectors
Usage: kcctl restart [-hV] [COMMAND]
Restarts some connectors or a task
  -h, --help      Show this help message and exit.
  -V, --version   Print version information and exit.
Commands:
  connector, connectors  Restarts the specified connector(s)
  task                   Restarts the specified connector or task
```

* Why does it mention the `task` sub-command, when displaying the help for the `connectors` sub-command? It seems that output above is actually for the `restart` command?
* Unrelated, but it seems the help message for `task` is wrong, it should just say "Restarts the specified task"?

gunnarmorling:
> Even though the -V flag is included in the usage summary for every command, it doesn't actually do anything when invoked with a command/subcommand (e.g., kcctl delete -V). When this happens, nothing is printed and the exit status is zero.

Hum, yeah, that's a bit weird.

> A bit of noise is added to the usage summary for every command. Commands that accepted a single argument or flag now have info on two additional flags (-h and -V), one of which is a no-op.

The -h seems justified, but -V not so much?

> Autocomplete support for the help command is completely wiped. We can probably restore it if that's a dealbreaker fro this PR.

Not a deal-breaker if it's _only_ for the help command. Maybe though we should get rid of it and _only_ support option 3 (-h, --help)?

C0urante:
I jumped the gun implementing this. Going to mark as a draft and continue the discussion on #363 until we've settled on an ideal syntax; you raised some good points there that deserve to be addressed before we spend time on code review. Thanks for the rapid turnaround though! 😄

C0urante:
Got rid of the extraneous `-V` flag. Took a little bit of legwork but hopefully the approach of adding a mixin to each command class isn't too brittle. If it is, I can try to look into more automatic/magical approaches that don't require per-command changes.

C0urante:
> Not a deal-breaker if it's only for the help command. Maybe though we should get rid of it and only support option 3 (-h, --help)?

Correct, it's only for the help command. I personally really like the `help <command> <subcommand>` syntax so I'll look into bring back autocomplete for it (hopefully including nested subcommands!).

C0urante:
> I'm confused by this one, it still shows the wrong one, no?

Sorry, missed this comment earlier! Was a copy+paste error on my part. I've updated the description with the correct output, after verifying that it accurately reflects what the CLI emits after being built with the latest commit for this PR.

C0urante:
> Unrelated, but it seems the help message for task is wrong, it should just say "Restarts the specified task"?

Yeah, good catch! I'll add it to this PR tomorrow if there are no objections.

gunnarmorling:
Sweet, this LGTM overall to me. IIUC, there's two remaining loose ends:

* Wrong message for `restart task` (+1 for adding this to this PR)
* Completion broken for help command

Anything else?

gunnarmorling (approved):

C0urante:
I think that's it!

With the latest commit, I fixed the `restart task` help description, and yesterday I "fixed" the `help` autocompletion. It's broken for nested subcommands, but works for top-level commands, which matches the logic on the current `main`. More info in the description.

I want to try to get autocomplete working for nested subcommands with `help` at some point, but don't think that should be a blocker. If you agree, I think we can merge this 🎉

gunnarmorling:
Sounds good, can you just log a follow-up issue for the nested subcommand help auto-completion?

gunnarmorling:
Applied. Thanks a lot, @C0urante!

C0urante:
Thanks Gunnar! Filed https://github.com/kcctl/kcctl/issues/370 as a follow-up for autocompletion of nested subcommands.

PR #392, opened 2023-10-02
#375 Add support for deleting offsets

mfvitale:
Closes: https://github.com/kcctl/kcctl/issues/375


mfvitale:
@C0urante can you please take a look?

src/main/java/org/kcctl/command/DeleteOffsetsCommand.java
```diff
@@ -0,0 +1,89 @@
...
+            return 1;
+        }
+
+        for (String connector : names) {
+            ConnectorOffsets offsets = kafkaConnectApi.deleteConnectorOffsets(connector);
+            // Display with JSON; the contents aren't very readable as raw text, and this output format
+            // plays nicely with the endpoint to patch connector offsets (you can use the output here as the
+            // body for requests to that endpoint)
+            String offsetsJson = mapper.writerWithDefaultPrettyPrinter().writeValueAsString(offsets);
+            spec.commandLine().getOut().println(offsetsJson);
```
C0urante, line 83: Is this correct? I thought the responses for altering/resetting offsets were different from the ones for reading them. The [KIP](https://cwiki.apache.org/confluence/display/KAFKA/KIP-875%3A+First-class+offsets+support+in+Kafka+Connect#KIP875:FirstclassoffsetssupportinKafkaConnect-Altering/resettingoffsets(response)) indicates that it should be something like this:

```json
{
  "message": "The offsets for this connector have been reset successfully"
}
```

or this:
```json
{
  "message": "The framework-managed offsets for this connector have been reset successfully. However, if this connector manages offsets externally, they will need to be manually reset in the system that the connector uses."
}
```

mfvitale: You are right. I'll make the changes that are needed. Good catch :wink:

src/test/java/org/kcctl/command/DeleteOffsetsCommandTest.java
```diff
@@ -0,0 +1,148 @@
...
+                    if (!isOffsetAvailable("delete-offsets-test1")) {
+                        // No offsets have been committed yet; wait a little longer for an offset commit to complete
+                        return false;
+                    }
+
+                    int exitCode = deleteContext.commandLine().execute("delete-offsets-test1");
+                    String errorMessage = deleteContext.error().toString();
+
+                    assertThat(exitCode).isEqualTo(CommandLine.ExitCode.SOFTWARE);
+                    assertThat(errorMessage).contains("Connectors must be in the STOPPED state before their offsets can be modified. This can be done for the specified connector by issuing a 'PUT' request to the '/connectors/delete-offsets-test1/stop' endpoint" );
```
C0urante, line 78: Does this section belong in the `await` block? Shouldn't it succeed immediately as long as at least one offset has been committed by the connector (and possibly even if no offsets have been committed)?

src/test/java/org/kcctl/command/DeleteOffsetsCommandTest.java
```diff
@@ -0,0 +1,148 @@
...
+                    }
+
+                    stopContext.runAndEnsureExitCodeOk("delete-offsets-test1");
+
+                    int exitCode = deleteContext.commandLine().execute("delete-offsets-test1");
+                    String output = deleteContext.output().toString();
+                    AlterResetOffsetsResponse response = mapper.readValue(output, AlterResetOffsetsResponse.class);
+
+                    assertThat(exitCode).isEqualTo(CommandLine.ExitCode.OK);
+                    assertThat(response.message()).isEqualTo("The offsets for this connector have been reset successfully");
```
C0urante, line 105: This also doesn't seem to belong in the `await` block. I think the flow we need is:

1. Wait for one or more offsets to be committed by the connector
2. Stop the connector
3. Wait for the connector to be stopped (i.e., no tasks left and the `Connector` is in the `STOPPED` state)
4. Delete offsets, which should succeed immediately by this point

mfvitale:
@C0urante thanks for the suggestions. I have addressed it and also I have updated the Debezium version to 2.4.0.Final to use the new `STOPPED` state. If you want to have this update on separate PR, I can do it.

C0urante (approved):
LGTM, thanks @mfvitale!

C0urante:
Regarding the bump in Debezium version--I'm fine with that in this PR 😄 

It looks like we can also start work on https://github.com/kcctl/kcctl/issues/346 now that we're on a version of Debezium that has native support for the `STOPPED` state! 🎉

Issue #399, opened 2024-01-09
Remove context


Xavier-Pisco:
I'm using kcctl and I wanted to do some tests with a specific account on my platform. I created a context with the new account and after I was done I wanted to delete that context, I didn't find a way to do this neither found an issue about this. 

For now I went to the .kcctl config file and just deleted the contexts I didn't want, but I think there should be a command, e.g. `kcctl config delete-context context-name`, for removing a context that was previously created and that's not needed anymore.

gunnarmorling:
Good catch, we should add that. PR welcome, if you're up for it.

Xavier-Pisco:
@gunnarmorling I created it, it's here https://github.com/kcctl/kcctl/pull/423

PR #423, opened 2024-01-19
#399 Remove context

Xavier-Pisco:
Add the command to allow for deletion of contexts that are not necessary anymore.

For precaution I chose to not allow the user to delete the current context, forcing them to change the context before deleting it.

### DIFF
--- a/src/main/java/org/kcctl/command/RemoveContextCommand.java
+++ b/src/main/java/org/kcctl/command/RemoveContextCommand.java
@@ -33,7 +33,7 @@ import java.util.Properties;
 import java.util.concurrent.Callable;
 
 @Command(name = "remove-context", description = "Removes a context from your contexts")
-public class RemoveContextCommand implements Runnable {
+public class RemoveContextCommand implements Callable<Integer> {
 
     @CommandLine.Mixin
     HelpMixin help;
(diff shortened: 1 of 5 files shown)


src/main/java/org/kcctl/command/RemoveContextCommand.java
```diff
@@ -0,0 +1,63 @@
...
+
+    @Parameters(index = "0", description = "Context name")
+    String contextName;
+
+    @Override
+    public void run() {
+        ConfigurationContext context = new ConfigurationContext();
+
+        if (context.getCurrentContextName().equals(contextName)) {
+            System.out.println("Set another context before removing the current context.");
```
gunnarmorling, line 49: It would be nice to return a non-0 code here (by implementing `Callable`). You can find an example in `DeleteOffsetsCommand`.

src/main/java/org/kcctl/command/RemoveContextCommand.java
```diff
@@ -0,0 +1,63 @@
...
+        }
+        else {
+            boolean success = context.removeContext(contextName);
+
+            if (success) {
+                System.out.println("Removed context '" + contextName + "'");
+            }
+            else {
+                System.out.println("Couldn't remove context; There was no context named '" + contextName + "'");
```
gunnarmorling, line 58: Same here.

gunnarmorling (commented):
LGTM overall. Two suggestions in regards to using a more meaningful return code for signalling error cases inline.

gunnarmorling:
Applied. Thanks a lot, @Xavier-Pisco!

PR #425, opened 2024-01-19
#424 Add DeleteOffsetsCommand as subcommand of DeleteCommand

### LINKED ISSUE #424: delete offsets command should be a sub-command of delete
`DeleteOffsetsCommand` needs to be hooked up as a sub-commend with `DeleteCommand`. I don't think it actually is invocable at all right now from the CLI?

C0urante:
Fixes #424


gunnarmorling:
Thx, @C0urante!

PR #431, opened 2024-01-22
🔧 Use /connectors/<CONNECTOR>/tasks for task configs

### LINKED ISSUE #430: Use /connectors/<CONNECTOR>/tasks for task configs
From Kafka 3.7, the `GET /connectors/{connector}/tasks-config` endpoint will be deprecated and it will be removed in Kafka 4.0 ([KIP-970](https://cwiki.apache.org/confluence/display/KAFKA/KIP-970%3A+Deprecate+and+remove+Connect%27s+redundant+task+configurations+endpoint)).

We should use the `GET /connectors/<CONNECTOR>/tasks` endpoint instead to retrieve task configs.

mimaison:
This fixes #430


gunnarmorling:
Applied, thanks a lot! Interesting one, also wasn't aware of that old endpoint. TIL :)

